PR #13, opened 2019-07-06
Inclusion of elliptical cross-section geometry.

BenjaminFraser:
Many thanks for providing this fantastic library for FEA various cross-sections - it's been extremely useful for some academic analysis I've been conducting over recent months!

Whilst using section-properties, I used it to analyse customised elliptical cross-sections with several inclusions of new classes to the sections.py file. As part of this I added a class for both solid and hollow elliptical sections, which was basically just some small modifications to the circular section classes.

If you think these are useful / appropriate, I've provided them in a pull request.

Kind regards,

Ben


sectionproperties/pre/sections.py
```diff
@@ -534,6 +534,142 @@ def __init__(self, d, n, shift=[0, 0]):
...
+                self.facets.append([i, 0])
+
+        self.shift_section()
+
+
+class Ehs(Geometry):
+    """Constructs an elliptical hollow section centered at the origin *(0, 0)*,
+    with outer vertical diameter *d_yo*, outer horizontal diameter *d_xo*, and 
+    thickness *t*, using *n* points to construct the inner and outer ellipses. 
+    Note that the thickness of a hollow ellipse does not stay constant all
```
Agent6-6-6: Hi Benjamin

I wonder if you can expand on this, it's my understanding that the perpendicular to surface through thickness is constant. I checked your code against the British EHS's and it agreed quite well so I assume your formulation is correct as it appears to create a section that has a constant through thickness based on plots. So the note saying the thickness is not constant was a little confusing to me. Can you elaborate on this if the through thickness is not constant. 

These were the only commercial EHS I could find properties for
[https://www.steelconstruction.info/images/e/eb/Celsius_355_NH_Technical_Guide_2018.pdf](https://www.steelconstruction.info/images/e/eb/Celsius_355_NH_Technical_Guide_2018.pdf)

robbievanleeuwen: Thanks Ben for the useful additions to the sections file! And thanks Agent for the proofing!

PR #18, opened 2019-11-12
Add nastran sections

JohnDN90:
This branch adds cross-section geometries commonly available in Nastran-based beam libraries. While these geometries could always be defined using `CustomSection` I thought it was be useful to have the Nastran geometries defined and parameterized in a way that is consistent with Nastran.

Python API documentation has been updated with information for the new sections.

A new example has been added demonstrating creation of one of the new sections.


JohnDN90:
I want to add some geometry checks to the new sections before you consider merging this pull request.  I'll get those changes pushed up later today, probably by 09:30, Sydney Australia time.

robbievanleeuwen:
Hi John,

Thanks for this pull request - only just getting the time to have a look at these, apologies for the delay! Quick question (I have never used NASTRAN before) - what is the relevance of the getStressPoints methods? 

Also, these methods seem to use the same input parameters (DIM1, ..., DIMX, shift) as those used to create the sections - would it make sense to store these variables as class variables and then refer to them in the getStressPoints functions? The getStressPoints method would then not need to take any parameters. For example:

```python
def __init__(self, DIM1, DIM2):
    self.DIM1 = DIM1
    self.DIM2 = DIM2

    # construct rest of method...

def getStressPoints(self):
    C = (0.5*self.DIM1-shift[0], 0.5*self.DIM2-shift[1])
    # etc.
```

Let me know your thoughts on this, maybe it doesn't make sense because I'm not understanding what the getStressPoints method is for. Also I'm happy to make these changes if you think they're relevant.

Robbie

Czarified:
Hi Robbie! I may be able to shed some light on this. In Nastran parlance, the stress recovery points are just the fiber locations on the section where Nastran will automatically evaluate and return data from (assuming stress results are requested). It of course has the stress function for the element, but data at these points it prints to the output file.

From the documentation:

> Ci, Di, Ei, Fi:  The y and z locations (i = 1 corresponds to y and i = 2 corresponds to z) in
> element coordinates relative to the shear center for the cross section located at
> X/XB. The values are fiber locations for stress data recovery.

So from the looks of it, each class method just automatically provides you the same points as the Nastran definition for each cross-section type.

JohnDN90:
> Also wondering if you could briefly explain the use case for this module? E.g. what is the main advantage of having these NASTRAN sections given a lot of these sections are included in the sections module?

Nastran is one of the most common FEA analysis packages and offers a wide variety of cross-sectional shapes for 1D elements.  Some of the shapes are complex enough that Nastran does it's own cross-sectional analysis (also using the Pilkey method) to determine section properties.  You can see many of the cross sections [here](https://knowledge.autodesk.com/support/nastran/learn-explore/caas/CloudHelp/cloudhelp/2018/ENU/NSTRN-Reference/files/GUID-40E4AA5C-1432-4516-999C-C82303877473-htm.html).  I've ran into several cases on my own when working with an input file (created by someone other than me) where the cross-sectional properties are defined using a PBARL or PBEAML card.  In these cases, dimensions of the cross-section are specified rather than the cross-sectional properties themselves.

This pull request would allow users to take the specified dimensions from a Nastran input file and enter them into section-properties to calculate the cross-sectional properties.  This is useful for debugging Nastran input files and also for "translating" them for use with other programs.  

There are multiple sections defined in Nastran which aren't currently pre-defined in section-properties.  Granted you could calculate these sections by specifying the cross-section using vertices, but I think having these sections pre-defined simplifies ease of use.  Some sections are repeated, for instance the BAR section in Nastran is the same as your RectangularSection.  I choose to repeat some of these cross-sections because I believe it makes the code more intuitive. If someone is using section-properties to figure out the properties of a Nastran section, I feel its more intuitive for them to look for a "BARSection" rather than a "RectangularSection", although they could use either.

> Quick question (I have never used NASTRAN before) - what is the relevance of the getStressPoints methods?

The stress point are the four points that Nastran uses to evaluate the stresses at in the cross-section.  These are represented by C, D, E, and F on the cross-section at the link above.  The coordinates C, D, E, and F are relative to the shear center of the cross-section in Nastran.  Here, I've made them relative to the origin, but you can still get them relative to the shear center by passing in the shear center as the `shift`.

> Also, these methods seem to use the same input parameters (DIM1, ..., DIMX, shift) as those used to create the sections - would it make sense to store these variables as class variables and then refer to them in the getStressPoints functions? The getStressPoints method would then not need to take any parameters.

That makes sense to me. We can go ahead and do that.

robbievanleeuwen:
Thanks for the detailed response! I have merged this branch into the master branch.

Issue #27, opened 2020-08-12
Plot Image Saving


Czarified:
**Is your feature request related to a problem? Please describe.**
No

**Describe the solution you'd like**
Switch at least the plot_centroids method to return the fig,ax objects so they can be properly saved and/or modified, regardless of execution method.

**Describe alternatives you've considered**
I've tried implementing a `plt.savefig(name)` command after calling `xsect.plot_centroids()`. This will work with `pause=False` if executed on the command line, however if executed somewhere like a Jupyter notebook, the "current figure" that `plt.savefig()` grabs is blank (even if pause is True). Some users are getting no plots while I'm getting correct plots. Online users always recommend saving prior to showing, or using the `plt.gcf()` to return a `Figure` object for referencing later.

This should be a simple fix by returning fig,ax in the plotting methods. Then, the figure is stored directly in memory and can be addressed by python to `fig.savefig()` instead.

It may break some existing implementations, if returned objects are not expected. That's why I wanted to bring up an issue rather than jumping straight into a new PR.

**Additional context**
An object oriented approach with matplotlib is better, in my opinion, than the "default" plt.plot and plt.show. It allows more exact control. Am I missing something obvious?

Czarified:
FYI, somehow I duplicated this issue. I think my internet connection was on the fritz and sent the command multiple times when trying to reestablish connection. I closed all the duplicates.

Czarified:
@robbievanleeuwen , any thoughts on this? I'd love to be told I'm wrong and missing some obvious way to just save the output plots without having to display them.

robbievanleeuwen:
Hi @Czarified, sorry dropped the ball on this.

Think everything makes sense and happy to implement the changes. Also implemented to other plotting methods.

Don't think changing the examples is necessary as you can ignore the returned values from the functions.

Czarified:
@robbievanleeuwen thanks for incorporating that, and expanding it! Any idea when the next release is planned, so these features make it over to the PyPI distribution?

PR #33, opened 2020-09-12
plot_centroids figure return

Czarified:
`CrossSection.plot_centroids()` now returns the plot figure and axes object.

Example scripts also updated to reflect this.

This PR should resolve Issue #27, allowing users more flexibility with the cross-section figure results. For example, adding points/arrows/circles to the figure, or saving the figure without relying on `plt.savefig(name)`, where the "current figure" isn't always correct.

### DIFF
--- a/sectionproperties/analysis/cross_section.py
+++ b/sectionproperties/analysis/cross_section.py
@@ -1177,6 +1177,8 @@ class CrossSection:
         # finish the plot
         post.finish_plot(ax, pause, title='Centroids')
 
+        return fig, ax
+
     def display_mesh_info(self):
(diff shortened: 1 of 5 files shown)


Czarified:
@robbievanleeuwen , since the dxf update seems to be taking a while, could I request a v1.0.8 packaged release with this change implemented? It would help me out to be able to pip install the library directly, rather than building the master branch myself, and I see no harm in including this small patch as really the only feature in 1.0.8.

Either way, it's no big deal! 😄

robbievanleeuwen:
Hi @Czarified ,

Apologies for the lack of response, I was recently in an accident and have had a bit of a slow recovery. I will try to have a look at this in the coming days :)

Robbie

PR #40, opened 2020-12-20
added dxf import works intermittently, then breaks

aegis1980:
Hi, I have been sitting on this for a while. 
Gave up on it because it works fine first time, and then randomly doesn't the second or third time around, running the same file (`example_dxfimport.py`). I have no idea why and no logs - just exits.  I think its a MeshPy thing.


robbievanleeuwen:
Hi @aegis1980,

Thanks for the code! Sorry for the slow take-up, life has been a bit up and down this year to say the least! There is another pull request attempting to achieve the same thing. I'm thinking I'll have a look at both of these over the next few days and maybe create a new branch combining both ideas so there can be a unified approach to loading dxf files.

Watch this space...

Robbie

aegis1980:
@robbievanleeuwen 
No worries! 
Just figured it better 'out there' rather than buried in another project (as it was). 
Great project btw. 
I stuck a demo dxf extrusion if you need one.

robbievanleeuwen:
Thanks again for the code - I believe this is being incorporated in #42.

Cheers,
Robbie

PR #41, opened 2020-12-21
add check_geometry function

### LINKED ISSUE #38: Plastic centroid/principal axis centroid not at elastic centroid for doubly symmetric section
**Describe the bug**
In the following code I would expect the plastic centroid to be at the origin (0, 0) and align with the elastic centroid. Yet it is reporting the plastic centroid below the origin. Section geometry centroid is shifted to 0,0 position.

Also, the horizontal principal axis is shown correctly at y=0, but the principal plastic axis centroid is not, being drawn below y=0.

The reported results agree with the plot in so far as the plastic centroid is not at 0,0 point.

Is this possibly a bug or maybe I'm missing something obvious here?

**To Reproduce**
Run code below and observe results

```
from sectionproperties.analysis.cross_section import CrossSection

mesh_area = 25  # max mesh size

d = 612  # I section depth
b_f = 229  # I-section flange width
t_f = 19.6  # I-section flange thickness
t_w = 11.9  # I-section web thickness
r_1 = 14  # I-section root radii
n_r = 50  # number of points considered around root radii

d_p = 650  # strengthening plate depth
b_p = 12  # strengthening plate thickness

# create constituent geometry
geometry1 = sections.ISection(d=d, b=b_f, t_f=t_f, t_w=t_w, r=r_1, n_r=n_r, shift=[-b_f / 2, -d / 2])
geometry2 = sections.RectangularSection(d=d_p, b=b_p, shift=[-b_f / 2 - b_p, -d_p / 2])
geometry3 = sections.RectangularSection(d=d_p, b=b_p, shift=[b_f / 2, -d_p / 2])

# create merged section
geometry = sections.MergedSection([geometry1, geometry2, geometry3])

# add holes
geometry.add_hole([-b_f / 2, 0])
geometry.add_hole([b_f / 2, 0])

# create mesh
mesh = geometry.create_mesh(mesh_sizes=[mesh_area, mesh_area, mesh_area])

# clean geometry
geometry.clean_geometry(verbose=True)

# create section
section = CrossSection(geometry, mesh)

# calculate results
section.calculate_plastic_properties()

# plot results
section.plot_mesh()
section.plot_centroids()

# display all results
# check https://sectionproperties.readthedocs.io/en/latest/rst/post.html for definitions
section.display_results(fmt='.3f')

# this will specifically display torsion and warping constants
J = section.get_j()
I_w = section.get_gamma()
print(f'J = {J}')
print(f'I_w = {I_w}')
```

**Expected behaviour**
For doubly symmetric section plastic and elastic centroid should align?

**Screenshots**
Note how principal plastic axis centroid does not align with the principal axes, and plastic centroid is shown below elastic centroid:-
Figure_1

**Additional context**
Using latest version 1.0.7

Spectre5:
Fixes #38 

For the issue code there, this produces the exception:

    sectionproperties.pre.pre.GeometryError: 2 points are too close to a facet:
    Hole @ [-114.5    0. ]
    Hole @ [114.5   0. ]

I have not extensively tested this code, so more testing would be advised before merging this.  And if the move to `shapely` gets finished, then it looks like there would be a bit of a "nicer" way to do this too based on what @Agent6-6-6 suggested in that issue.  But this code is fully vectorized using numpy, so no python loops.

This new `check_geometry` function does 2 checks:

- Make sure that no facets have the same starting and ending point (length of zero) --- I assume that is invalid, but please confirm this
- Checks that no holes nor control points are within `tolerance` of a facet

### DIFF
--- a/sectionproperties/pre/pre.py
+++ b/sectionproperties/pre/pre.py
@@ -535,7 +535,7 @@ def check_geometry(points, facets, holes, control_points, atol=1.0e-8):
     check_points = np.array(holes + control_points)
 
     # select the starting and ending points of each facet
-    facet_points = points[facets]
+    facet_points = points[facets, :]
     start = facet_points[:, 0, :]
     end = facet_points[:, 1, :]
 
@@ -580,9 +580,9 @@ def check_geometry(points, facets, holes, control_points, atol=1.0e-8):
     minimum_distances = np.hypot(clamped_distances, perpendicular_distances)
 
     # make sure that the closest distance is not within the allowed tolerance
-    if np.min(minimum_distances) < tolerance:
+    if np.min(minimum_distances) < atol:
         # select the offending points and point type, then raise an exception with that information
-        distance_check = (minimum_distances < tolerance).sum(axis=0)
+        distance_check = (minimum_distances < atol).sum(axis=0)
         point_types = np.array(['Hole'] * len(holes) + ['Control Point'] * len(control_points))
         offending_types = point_types[distance_check > 0]
         offending_coords = check_points[distance_check > 0]
--- a/sectionproperties/pre/pre.py
+++ b/sectionproperties/pre/pre.py
@@ -511,7 +511,7 @@ class GeometryCleaner:
             self.geometry.facets.pop(idx)
 
 
-def check_geometry(points, facets, holes, control_points, tolerance=1.0e-7):
+def check_geometry(points, facets, holes, control_points, atol=1.0e-8):
     """Check that holes and control points do not lie on a facet line.
 
     :param points: List of points *(x, y)* defining the vertices of the cross-section
@@ -524,8 +524,8 @@ def check_geometry(points, facets, holes, control_points, tolerance=1.0e-7):
     :param control_points: A list of points *(x, y)* that define different regions of the
         cross-section. A control point is an arbitrary point within a region enclosed by facets.
     :type control_points: list[list[float, float]]
-    :param tolerance: minimum permissable distance from facet
-    :type tolerance: float
+    :param atol: minimum permissable point distance from any section facet
+    :type atol: float
 
     General method was adapted from:
     https://stackoverflow.com/a/54442561
@@ -543,8 +543,7 @@ def check_geometry(points, facets, holes, control_points, tolerance=1.0e-7):
     facet_vectors = end - start
     lengths = np.sqrt(np.einsum('ij,ij->i', facet_vectors, facet_vectors))
     if np.any(lengths == 0.0):
-        # if a length is zero, then a facet had the same starting and ending
-        # point
+        # if a length is zero, then a facet had the same starting and ending point
         facet_indices = np.nonzero(lengths == 0)[0]
         offending_facets = facet_points[facet_indices]
         facet_point_indices = np.array(facets)[facet_indices]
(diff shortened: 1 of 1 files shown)


sectionproperties/pre/pre.py
```diff
@@ -505,7 +511,100 @@ def remove_duplicate_facets(self):
...
+    facet_points = points[facets]
+    start = facet_points[:, 0, :]
+    end = facet_points[:, 1, :]
+
+    # create the vectors for each facet and normalize them to unit vectors
+    facet_vectors = end - start
+    lengths = np.sqrt(np.einsum('ij,ij->i', facet_vectors, facet_vectors))
+    if np.any(lengths == 0.0):
+        # if a length is zero, then a facet had the same starting and ending
+        # point
```
robbievanleeuwen, line 547: I'm using a line length of 100 characters. This could probably be placed on line 546 for prettiness (I'm a bit OCD haha)

sectionproperties/pre/pre.py
```diff
@@ -505,7 +511,100 @@ def remove_duplicate_facets(self):
...
+    )
+
+    # calculate the cross product to get the perpendicular distance
+    perpendicular_distances = np.cross(vec1, unit_vectors.reshape(-1, 1, 2))
+
+    # get the minimum clamped distance
+    minimum_distances = np.hypot(clamped_distances, perpendicular_distances)
+
+    # make sure that the closest distance is not within the allowed tolerance
+    if np.min(minimum_distances) < tolerance:
```
robbievanleeuwen, line 589: I think it would be best to use an absolute and relative tolerance here (e.g. see the zip_points() method).

robbievanleeuwen: Actually on second thoughts we could probably stay with absolute tolerance as zipping nodes is a bit of a different process where the user might have very large dimensions and what nodes zipped which are 'relatively' close. I think probably in this case we can get away with just an absolute tolerance as currently programmed. Maybe change tolerance to 1e-8 to keep consistency with the other methods?

robbievanleeuwen (commented):
This looks awesome, thanks for this! I love the pythonic way you've done things! Pretty happy to merge this,  let me know your thoughts on the tolerance business and then we'll get it incorporated and close the relevant issue!

Thanks again!

sectionproperties/pre/pre.py
Spectre5: and I'm OCD with using `black` but with line length of 79!  Haha.  But of course I should follow whatever the code style is of the library I'm committing too, I just wasn't sure what the section-properties default width was.  I'll update to make to 100 for this PR.

sectionproperties/pre/pre.py
Spectre5: Yup, I should have looked around for other tolerance values in the code - I agree that it should be consistent.  Also, I agree that only an absolute tolerance is needed here, I think.  Maybe I'll rename it to be more clear too or follow whatever other tolerance names that are in the code base.

Spectre5:
Ok, I pushed a change to address your comments.  Let me know if you have any other comments!

Spectre5:
Oh, I didn't realize that this failed the CI tests, I thought it was passing when I last committed!  I'll look into this and fix it later today!

Spectre5:
ok, pushed a fix!

PR #45, opened 2021-01-06
spelling corrections

Spectre5:
Entire PR is only spelling corrections and there should be no functional change to the code itself and the majority of the corrections are in comments and a few printed strings.

- There are one or two inner function names changed (for example, `get_prinicipal_points` to `get_principal_points`)
- I changed two instances of `’` to `'`
- I left the Aussie "s" spelling on words like discretised (vs the US discretized), ~but I did change artefacts to artifacts as I've never seen it spelled artefacts before~


docs/source/rst/examples.rst
```diff
@@ -411,7 +411,7 @@ class. The order of the geometry objects in the list that is passed into the con
 order relates to specifying mesh sizes and material properties.
 
 Once the geometry has been merged, it is vital to clean the geometry to remove
-any artefacts that may impede the meshing algorithm. A mesh is created with a mesh
+any artifacts that may impede the meshing algorithm. A mesh is created with a mesh
```
robbievanleeuwen, line 414: artefact is British English which is what I use :smile:

docs/source/rst/geom_mesh.rst
```diff
@@ -181,7 +181,7 @@ Cleaning the Geometry
 ---------------------
 
 When creating a merged section often there are overlapping facets or duplicate nodes. These
-geometry artefacts can cause difficulty for the meshing algorithm. It is therefore recommended to
+geometry artifacts can cause difficulty for the meshing algorithm. It is therefore recommended to
```
robbievanleeuwen, line 184: see above.

robbievanleeuwen (commented):
Awesome work! Looks like I've made a few copy paste errors (assembing haha). Thanks for picking these up! Only comment is regarding artefact which I believe to be British English.

Thanks!

docs/source/rst/examples.rst
Spectre5: changed back.

docs/source/rst/geom_mesh.rst
Spectre5: changed back.

Spectre5:
Ok, I changed it back!  I know about "our" endings like colour and "s" vs "z" differences, but artefact is a new one for me :)

PR #47, opened 2021-01-08
some linting cleanup

Spectre5:
As more people become interested in helping grow this project, it would be a good idea to set some clear "requirements" for code style and linting.  I'd recommend that `black` be used for formatting, `pylint` and/or `flake8` for linting, and potentially `isort` for sorting imports (with black compatibility mode).  Check with all of these can be incorporated into the CI so that non-conforming code is rejected.  I can help with some of that setup, if you're interested in that, to make the libraries play nice together.  You could even include a sort of "build" script that would setup the virtual environment, install the dev dependencies, and could be run to do a code check locally to be run before committing.

Anyway, this PR contains some changes based on using pylint and flake8.  I disabled some errors that I don't suspect you'd want to be changed, but even so this does not fix all of the messages.  If you like these changes, then I'll plan to do a few others too.  Most of the changes in this PR boil down to one of:

-  Don't use `str` not `exit` as variable names, which are built in python names
-  Removed unused variables, for example on a line like `(fig, ax) = plt.subplots()` where only the ax is used, I've replaced `fig` with `_` which is the de facto standard to indicate that we don't care about that return value
-  Removed argument `ax` from `setup_plot` (not used in function)
-  Converted a handful of method to use the staticmethod decorator since they didn't use `self` at all
-  Changed two of the plotting functions to return `None, None` if `ax_supplied` is true.  Best practice is that a function that returns something should always return something.
- Simplified a number of if/else loops
-  Removed lists as default values to methods (dangerous if the function is called multiple times!)

And perhaps the biggest change here is that I've changed `GeometryCleaner` to be `GeometryCleanerMixin` for use as a mixin to the Geometry class.  The way `clean_geometry` was handled before was rather weird where the returned geometry object gets re-assigned back to `self`.  After this PR, the methods from the geometry cleaner could just be moved now straight into the `Geometry` class, but perhaps it is nicer to keep it as a mixin to clearly differentiate the cleaning functions from the rest of the class.


robbievanleeuwen:
This is great! Keen to take this forward and formalise.

I'm not super familiar with `black` or incorporating into the CI - do you need to specify some formatting rules (e.g. line widths etc.) and tell the CI what these rules are? And how is this information usually conveyed to potential contributors?

Happy to be taken to school on this one :)

Czarified:
My (limited) experience with `black` is that there is really nothing to configure or setup. It does it one way, and that's it. For libraries that I know of using it, they have a badge on their readme and a short blurb/link saying they use black, and whether CI will automatically do it.

There are some things in my initial testing that it didn't handle well, like long raw strings. For the most part it's just plug and play.

Spectre5:
`black` is intentionally very opinionated and has very few options as its intention is to minimize discussion on what the "best" formatting is and allow you to instead focus on the code itself.  I've used it extensively for every project and script I've written for years now and it is fantastic, in my opinion.  Admittedly, you'll fight it the first couple of times you use it since it may format something "not the way I want it"....but after you accept it, you get used to it and learn to really appreciate it.  Typically I have my editor to just format the code each time or save, or else regularly format it myself.  Then you can just kind of code and ignore most all of the formatting, then like black fix it all.

I personally do use two of the settings it does offer.  One is the line length (so 100 for this project) and another I always use is to "skip string normalization" as I personally dislike double quotes.  By default, it (generally) converts all single quotes to double quotes, but with the skip option it just leaves them as-is, so they stay single (`'`) or double quoted quoted (`"`) based on whatever you type in the code.

There is 1 time I've seen that it sometimes doesn't respect your line length, and that is with really long strings.  It tries to respect it, but it it cannot, then it lets the line be too long.  Image, for example, a string that is 150 characters long.  Obviously it cannot make that string fit within a 100 character length line.  So in this case, you can accept the long line length or you can manually wrap it yourself (what I do).  To manually wrap it, you can just use pythons implicit string concatenation within tuples.  The `black` project is in the early stages of implementing automatic string wrapping too, but it is actually a very tricky problem with many edge cases, so it isn't in their released versions as of now, but is available for early testing.

As for telling people about it (black and any other tools you choose to enforce, like `flake8`, etc), there are generally a few ways:

- Often a project uses a badge at the top of the README as @Czarified mentioned.
- Most projects will have a contribution guide, either as a section to the README or as a separate CONTRIBUTING.md (or rst) file that the README links to (common if the contribution guide is on the longer side).
- Through enforcement in the CI, so that if a PR is opened that doesn't meet your code standards, the contributor knows right away as it fails CI.
- You might have some sort of a "build" or "run" script (or even a Makefile) at the top level of the directory which is likely to catch the eye of most contributors --- and then this script can run the format and linting tools.
- More experienced users will often check the settings in the project for various commonly used format/lint tools (`pyproject.toml` for `black`, others for other tools).

I'll take a stab at a pretty basic implementation with a few common tools and open a new PR and we can all discuss it further there, including if we want to keep or remove each tool.  And of course it can always change in the future too.

robbievanleeuwen:
Thanks for the rundown! If either of you are interested I think this would be an excellent addition to the project!

As this PR has already been merged (and buried a little bit in the gihub interface) I've created a new discussion at #52.

PR #63, opened 2021-03-06
Save figures to file

Spectre5:
This makes a few changes.  The code was refactored a bit to use a context manager for the plotting, which has several benefits:

- All plot functions can accept a title now, in case the user wants a unique one (perhaps to include a load case name that is being analyzed)
- I've added a `filename` parameter.  This can be used to save directly to a file without displaying the plot on the screen.  This is useful if you're analyzing many sections and don't want all of the plots to be shown on the screen.  So you can do, for example, if you do `geometry.plot_geometry(filename='geometry.png')`, then the image will be saved to file without showing on the screen.
- Add a `render` option.  If a filename is provided, it gets set to False automatically.  But it is also provided separate as a user that is adding a section to another axes may not want the plot to render on the screen immediately, so they can turn off rendering and then display it on the screen when they choose.
- The plot functions accept **kwargs that get sent to the plotting context, which cleans up some of the functions.  Also, any keywords that it doesn't use get sent to the `subplots` matplotlib function.  This is nice as a user can then pass extra arguments.  for example, `geometry.plot_geometry(figsize=(10,10))` could be used to create a larger figure size than the matplotlib default.


Spectre5:
I've now also added the `axis_index` parameter so that multiple subplots is supported.  So you can, for example, do this:

```python
from sectionproperties.analysis.cross_section import CrossSection
import matplotlib.pyplot as plt

geometry = sections.CircularSection(d=50, n=64)
fig, _ = geometry.plot_geometry(nrows=2, ncols=1, figsize=(8,10), render=False)

mesh = geometry.create_mesh(mesh_sizes=[2.5])

section = CrossSection(geometry, mesh)
section.display_mesh_info()
ax2 = fig.axes[-1]
section.plot_mesh(ax=ax2)
plt.show()
```

And get the geometry and mesh in 1 figure:

Figure_1

Spectre5:
As for the failing CI....it looks like matplotlib has bumped the requirement for the latest version of 3.4:

```
Beginning with Matplotlib 3.4, Python 3.7 or above is required.
You are using Python 3.6.7.
```

So we need to either require matplotlib < 3.4 in the requirements or else drop support for python 3.6 ourselves (my preference).  I can include dropping that support in this PR if you want @robbievanleeuwen.  But I think all PR will fail until one of these two actions occur though.

Spectre5:
Note that numpy also dropped support for python 3.6 with the newest version of 1.20.  So if we need to limit it too if we want to support python 3.6 still.  One option would be to use `python_version` to limit the numpy and matplotlib versions for python 3.6 but allow newer ones for 3.7+.  I'd still rather just drop 3.6 myself.

Agent6-6-6:
> And get the geometry and mesh in 1 figure:
> 
> Figure_1

Great work on this!

Just a thought, I kind of think it would be good to have a built in option to condense several common plots into one in some standardised layout like you've shown. Just saves user having to setup code to do something logical like you've shown. 

For example the stresses and stress vectors having an option to return all the stress plots for various actions in one figure kind of makes sense so you can see all the relevant info in one plot?

For example geometry, centroids and mesh in one plot in 3x1 arrangement or similar?

Spectre5:
I agree - that is a good idea and I had a similar though when I implemented this.  I can probably put something together like that in a future PR.  I'd have to give the implementation a little thought, but I think it would best be in a new "utilities" module where you'd call the function and provide a section/analysis/whatever and it would then generate the plot.  I think that would be a clean way to implement it and provides a natural way to extend it in the future too.

Czarified:
I like these changes a lot, and I'm also in favor of dropping 3.6 and having the min python req. at 3.7.

I implemented the original fig, ax return because I needed a way to save the figures and automatically pull them into report builds, not even showing them is even better for me! 😄 This seems like a thorough, stable, but flexible way to control figure outputs.

I have not reviewed the code, but I may have some time today, just to get any additional comments for you. Thanks Spectre5!

Spectre5:
That'd be great!

Czarified (commented):
I have not tested this code locally, but all the changes make sense to me. I see no issues.

Issue #21, opened 2020-04-18
Overhaul pre-processor to include shapely


robbievanleeuwen:
**Is your feature request related to a problem? Please describe.**
No specific problems, but would increase functionality and simplify a fair bit of the code.

**Describe the solution you'd like**
Use shapely in all it's glory.

**Describe alternatives you've considered**
Writing the code yourself - tried that, no fun.

**Additional context**
Suggesting in issue #9 by Agent6-6-6.

aegis1980:
Hi, 
Am sure I saw a 'DWG import' todo somewhere - this is more related to that, but I am using shapely. 
A WIP but (as part of another project) I made this: https://github.com/aegis1980/cad_to_shapely. I am having a few issues with making is work on PyPi (imports mostly: Python not my goto) but at some point it will be a pip-able thing.
Currently only DXF (I ve written code for IGES and SVG - need to add it).
Anyway, [this gist](https://gist.github.com/aegis1980/a6cbc474af6d6f979f0dcb5ed4047e19) imports a (Rhino3D generated DXF) aluminium extrusion (file attached) and spins it around for sectionproperties processing.
Need to check I get the same answers as I get using Strand7 so unverified.
Good work BTW ;) And a strong antipodean influence too.
![Figure_1](https://user-images.githubusercontent.com/4284988/86887884-65a24f80-c14d-11ea-9c02-578f56b3c8bd.png)
[section_holes_complex.zip](https://github.com/robbievanleeuwen/section-properties/files/4888992/section_holes_complex.zip)

connorferster:
Hi @robbievanleeuwen,

I have some time before January and I would like to work on this. I am experienced with shapely and have previously started looking into how to integrate it into section-properties. 

I will make a fork and start working on a branch so you can review the progress. As I go, let me know if you would like to advise me further or chat about the direction it's going. 

Thanks for this excellent library! I think using shapely will be an excellent addition to it.

robbievanleeuwen:
Hi @connorferster,

I initially had a few tries at implementing shapely but ran into a few hurdles and didn't delve into the docs deep enough to pull through. If you think you could make it work I think it would be a fantastic addition and potentially open up a few other possibilities with _sectionproperties_, so I'm more than happy for you to give it a try!

Give me a shout if you have any questions or would like to discuss anything 👍 

Cheers,
Robbie

connorferster:
[...]

PR #66, opened 2021-04-20
Shapely preprocessor

connorferster:
Here is the pull request for review of the shapely pre-processor. This introduces several breaking changes to the pre-processor and is intended to commence as v2.0.0 of sectionproperties.


sectionproperties/pre/sections.py
```diff
@@ -1,259 +1,568 @@
...
+        return
 
-        :return: Object containing generated mesh data
-        :rtype: :class:`meshpy.triangle.MeshInfo`
+    def compile_geometry(self):  # Alias
+        # pass
+        self.create_facets_and_control_points()
 
-        :raises AssertionError: If the number of mesh sizes does not match the number of regions
+    def create_mesh(self, mesh_size: float):
```
Agent6-6-6, line 184: this should remain as `mesh_sizes`, see my post in issue #21, as it breaks single geometry vs compound geometry

sectionproperties/pre/sections.py
```diff
@@ -1,259 +1,568 @@
...
+            new_y = abs_y if abs_y else current_y + dy
+            current_points[point_idx] = (new_x, new_y)
+
+        new_geom = Geometry.from_points(
+            current_points, current_facets, current_holes, current_control_points,
+        )
+        return new_geom
 
-    def plot_geometry(self, ax=None, pause=True, labels=False, perimeter=False):
+    def plot_geometry(
```
Agent6-6-6, line 553: perimeter = True does not seem to trigger the linewidth = 3 line below?

connorferster: This is related to the calculation of the exterior perimeter. If the shapes are not disjoint, I can calculate (and plot) the exterior perimeter. If shapes are disjoint, then `CompoundGeometry.calculate_perimeter()` returns -1.0. In the instance of shapes being disjoint, what are your thoughts on the behaviour of `CompoundGeometry.plot_geometry(perimeter=True)`?

Agent6-6-6: If disjoint I'd love to see it reflected on the plot somehow as a warning to the user, for example a red perimeter around the disjoint object(s) as you'd have several external perimeters?

Agent6-6-6: I wonder since the results are invalid if you have disjoint geometry if the code should simply raise an exception and exit? Currently you can get the results either way, which if you don't realise by plotting geometry could lead to incorrect results. Perhaps you could have an optional option to halt_on_disjoint_geometry or similar as a thought?

Spectre5:
A lot of work here!  Very cool.  I hope to look over the changes myself too, but it'll probably be almost a week before I'm able to contribute anything.  Glad to see that you've continued to develop this.

sectionproperties/pre/sections.py
connorferster: I don't think the results are invalid for disjoint geometry. I just did a test against a handcalc for two separate but equal rectangles and the result is correct (see below). Tell me more about this experience you had with incorrect results. 

To be honest, I _could_ increase the lineweight for disjoint geometry, too. It's just more leg-work because it will require at least three or four helper functions to do (iterate over Geometry in unionized CompoundGeometry, accumulate exterior rings, create facet lists for each exterior ring, and pass those in sequence to `.plot_geometry()`) and I am being a baby about doing it right now :)

image

Agent6-6-6: I should have said some of the properties are invalid. Robbie could probably add more to the reasons, but torsion and warping functions come to mind. It was in the docs previously for example that the results are only valid if you have no disjoint geometry. It's also come up in standard responses from Robbie in some prior issues. 

Understood on highlighting disjoint geometry. To be honest shapely version is much less error prone in this respect. I've had instance where I've put elements in contact but forgot to clean_geometry previously and they haven't meshed properly. At least with shapely version this is no longer an issue are the merging/cleaning is all handled by shapely seemlessly.

sectionproperties/pre/sections.py
connorferster: Confirmed. Fixed!

sectionproperties/pre/sections.py
connorferster: Ah! Thank you for the details. That all makes sense.

Interestingly, if you create two shapes and put them in contact as a CompoundGeometry, shapely will turn the svg image red because they are invalid for shapely (but valid for sectionproperties). If they stay green, then they are not touching. So there is something like this built in to the new preprocessor.

Regarding the perimeter arg in plot_geometry(), I will perhaps remove it as an argument from CompundGeometry for now until that is implemented.

connorferster: I should add that invalid geometry in shapely only affects certain shapely operations which will raise an error when given invalid geometry to operate on. We have not been using such operations in sectionproperties.

Agent6-6-6: > Regarding the perimeter arg in plot_geometry(), I will perhaps remove it as an argument from CompundGeometry for now until that is implemented.

Yeah makes sense.

sectionproperties/analysis/cross_section.py
```diff
@@ -190,11 +179,14 @@ def init():
             self.mesh_elements = elements
             self.mesh_attributes = attributes
 
+            # log.log(level=logging.DEBUG, msg=f"Section:\n")
+            # log.log(level=logging.DEBUG, msg=f"Tri6 Elements: {self.elements}")
+
             # initialise class storing section properties
             self.section_props = SectionProperties()
 
         if time_info:
```
Agent6-6-6, line 188: @connorferster, I think this should this be `self.time_info`, not `time_info`?

Agent6-6-6: Also on line 182 above this comment, is this commented logging code required

connorferster: @Agent6-6-6 Good idea! If we change these args to default to self.time_info then it could streamline things nicely.

connorferster: Done!

sectionproperties/analysis/cross_section.py
```diff
@@ -834,16 +819,14 @@ def calc_plastic():
             plastic_section = PlasticSection(self.geometry, debug)
             plastic_section.calculate_plastic_properties(self, verbose)
 
-        if time_info:
+        if self.time_info:
             text = "--Calculating plastic properties...\n"
```
Agent6-6-6, line 823: To be consistent with other time outputs remove the addition of the new line at the end `\n`?

sectionproperties/analysis/cross_section.py
```diff
@@ -834,16 +819,14 @@ def calc_plastic():
...
             solver.function_timer(text, calc_plastic)
             print("")
```
Agent6-6-6, line 825: Consider removing these print statements and instead add `\n` to the end of the solver.function_timer function print statement where the time is printed. 

This would add a blank line between any time outputs and get rid of all the `print("")` lines? Currently it's a bit inconsistent, where some `self.time_info` statements have a blank line added using `print("")` and others do not. I think this would tidy up output nicely.

sectionproperties/analysis/solver.py
```diff
@@ -126,7 +126,6 @@ def function_timer(text, function, *args):
     result = function(*args)
 
     if text != "":
-        print("----completed in {0:.6f} seconds---".format(
-            time.time() - start_time))
+        print("----completed in {0:.6f} seconds---".format(time.time() - start_time))
```
Agent6-6-6, line 129:
```suggestion
        print("----completed in {0:.6f} seconds---\n".format(time.time() - start_time))
```

Suggested along with removing `print("")` lines in all `self.time_info` checks? Add new line at end of this statement being printed to terminal to put line between all time outputs.

sectionproperties/analysis/fea.py
```diff
@@ -58,19 +64,30 @@ def geometric_properties(self):
         for gp in gps:
             # determine shape function, shape function derivative and jacobian
             (N, _, j) = shape_function(self.coords, gp)
+            # log.log(level=logging.DEBUG, msg=f"N: {N}\nj: {j}")
```
Agent6-6-6, line 67: Remove temporary log/commented out code?

sectionproperties/analysis/fea.py
```diff
@@ -321,6 +383,8 @@ def plastic_properties(self, u, p):
             whether or not the element is above the line
         :rtype: tuple(float, float, float, float, bool)
         """
+        # import warnings
```
Agent6-6-6, line 386: Remove temporary log/commented out code?

sectionproperties/analysis/fea.py
```diff
@@ -343,15 +407,43 @@ def plastic_properties(self, u, p):
             force += gp[0] * j * self.material.yield_strength
 
         # calculate element centroid
+        # try:
```
Agent6-6-6, line 410: Can this commented out try statement be removed?

sectionproperties/analysis/solver.py
Agent6-6-6: or
```suggestion
        print(f"----completed in {time.time() - start_time:.6f} seconds---\n")
```

sectionproperties/analysis/cross_section.py
```diff
@@ -1049,41 +1118,53 @@ def plot_mesh(self, ax=None, pause=True, alpha=1, materials=False, mask=None):
...
                 if i == 0 or material not in self.materials[0:i]:
                     # add the material colour and name to the legend list
-                    legend_list.append(mpatches.Patch(color=material.color, label=material.name))
+                    legend_list.append(
+                        mpatches.Patch(color=material.color, label=material.name)
+                    )
 
             cmap = ListedColormap(color_array)  # custom colormap
-            c = np.arange(len(color_array))  # indices of elements
+            # c = np.arange(len(color_array))  # indices of elements
```
Agent6-6-6, line 1151: Should this be deleted, because you're populating variable c above in a different manner with the changes?

sectionproperties/analysis/cross_section.py
```diff
@@ -1309,21 +1404,25 @@ def get_ig(self):
...
+        return (
+            self.section_props.ixx_g,
+            self.section_props.iyy_g,
+            self.section_props.ixy_g,
+        )
 
     def get_c(self):
         """
-        :return: Elastic centroid *(cx, cy)*
+        :return: Elastic centroid *()*
```
Agent6-6-6, line 1420: I think this should be discarded/reverted, `:return: Elastic centroid *(cx, cy)*` seems correct here?

sectionproperties/analysis/cross_section.py
```diff
@@ -1758,109 +1872,19 @@ class PlasticSection:
...
+        self, geom: Union[sections.Geometry, sections.CompoundGeometry], debug: bool
+    ):
         """Inits the PlasticSection class."""
-
-        # make a deepcopy of the geometry & materials so that we can modify it
-        self.geometry = copy.deepcopy(geometry)
-        self.materials = copy.deepcopy(materials)
+        self.geometry = geom.align_center()
+        self.geometry.compile_geometry()
         self.debug = debug
```
Agent6-6-6, line 1881: is this actually used anymore given the way the section is no longer meshed at each iteration with shapely version? 

I could not find anywhere where it is utilised, so I assume it can be removed as an argument and the associate doc for the argument?

sectionproperties/analysis/cross_section.py
connorferster: I think I should just change the other outputs :) They are printing an empty string to create a new line. A new line character is more explicit. You are right, they should be consistent!

sectionproperties/analysis/cross_section.py
connorferster: Yes, I agree. I will do this instead.

sectionproperties/analysis/fea.py
connorferster: Darn it! I did a Ctrl-F in cross_sections.py yesterday but not in fea.py. Will do!

sectionproperties/analysis/fea.py
connorferster: Yep

sectionproperties/analysis/cross_section.py
connorferster: Yep!

sectionproperties/pre/sections.py
```diff
@@ -1,259 +1,573 @@
...
+        if on in ["top", "bottom"]:
+            arg = "y_offset"
+        kwargs = {arg: offset}
+        new_geom = self.shift_section(**kwargs)
+        return new_geom
 
-            return np.arctan2(pt1[1] - pt2[1], pt1[0] - pt2[0])
+    def align_center(self, align_to: Optional[Geometry] = None):
+        """
+        Returns a new Geometry object, tranlsated in both x and y, so that the
```
Agent6-6-6, line 294: small spelling mistake 👍 
```suggestion
        Returns a new Geometry object, translated in both x and y, so that the
```

sectionproperties/pre/sections.py
```diff
@@ -2002,129 +2544,115 @@ class PolygonSection(Geometry):
 
         Mesh generated from the above geometry.
     """
+    outer_points = []
+    inner_points = []
 
-    def __init__(self, d, t, n_sides, r_in=0, n_r=1, rot=0, shift=[0, 0]):
-        """Inits the PolygonSection class."""
+    if n_sides < 3:
+        msg = "n_sides required to be greater than 3 for PolygonSection class"
```
Agent6-6-6, line 2551: The exception message here still refers to PolygonSection class, which is now turned into polygon_section geometry function

Change to something like this?
```suggestion
        msg = "n_sides required to be greater than 3 for polygon section type"
```

sectionproperties/pre/sections.py
```diff
@@ -2002,129 +2544,115 @@ class PolygonSection(Geometry):
...
-        # specify a hole in the centre of the Polygon section
-        self.holes = [[0, 0]]
+        # append the current temporary points to the temporary points list
+        outer_base_points.append([x_outer, y_outer])
+        inner_base_points.append([x_inner, y_inner])
 
-        # start at bottom face, constructing one corner radii, then rotate by initial rotation +
-        # alpha and repeat for n_side number of times to form full section perimeter
+    # if radii merged to circle with an outer diameter of a_out then skip last point as causes
+    # overlapping end points which causes meshing issues if geometry is not cleaned by user
```
Agent6-6-6, line 2613: This might no longer be required with shapely now handling the geometry rather than relying on prior clean_geometry action.

Have not tested polygon_section yet though.

sectionproperties/analysis/cross_section.py
connorferster: Are you referring to `self.debug`? It is something that Robbie had included in the code for trouble shooting. When I redid the implementation, all of the methods that took a `debug` argument were replaced. So, yeah, we should definitely remove it :)

sectionproperties/pre/sections.py
connorferster: Fixed!

sectionproperties/pre/sections.py
connorferster: Fixed!

sectionproperties/pre/sections.py
connorferster: Removed!

sectionproperties/analysis/cross_section.py
connorferster: Removed!

sectionproperties/analysis/fea.py
connorferster: Removed!

sectionproperties/analysis/fea.py
connorferster: Removed!

sectionproperties/analysis/fea.py
connorferster: Removed!

sectionproperties/analysis/cross_section.py
connorferster: Agreed. Done!

connorferster:
Thank you @Agent6-6-6 for all of the edits and suggestions. That has been really helpful! We have done a lot of work cleaning up inconsistencies in the code! Like 15 issues? So awesome.

@Spectre5 Would love to get your input, also!

sectionproperties/analysis/cross_section.py
Agent6-6-6: Yeah was referring to self.debug, as didn't seem to be used anymore with the new methods of solving for plastic axis 

(github is a little unclear on what line was commented on sometimes to me as well!)

connorferster:
Excellent. Got it!

sectionproperties/pre/sections.py
```diff
@@ -470,11 +477,11 @@ def split_section(
 
         return (top_right_geoms, bottom_left_geoms)
 
-    def offset_section_perimeter(self, amount: float = 0, resolution: float = 12):
+    def offset_perimeter(self, amount: float = 0, resolution: float = 12):
```
Agent6-6-6, line 480: @connorferster, this method seems to fail if you have a compound geometry you want to offset without any specifically defined material(s)?

Have not checked with materials if it works.

Can you please check and see if you're seeing the same thing?

Agent6-6-6: Actually I think it's more to do with the fact that materials have not been implemented in CompoundGeometry yet?

```
AttributeError: 'CompoundGeometry' object has no attribute 'material'
```

sectionproperties/pre/sections.py
```diff
@@ -293,224 +622,757 @@ def plot_geometry(self, ax=None, pause=True, labels=False, perimeter=False):
...
-    :param control_points: A list of points *(x, y)* that define different regions of the
-        cross-section. A control point is an arbitrary point within a region enclosed by facets.
-    :type control_points: list[list[float, float]]
-    :param shift: Vector that shifts the cross-section by *(x, y)*
-    :type shift: list[float, float]
-    :param perimeter: List of facet indices defining the perimeter of the cross-section
-    :vartype perimeter: list[int]
+            geometry_1 = sections.i_section(d=203, b=133, t_f=7.8, t_w=5.8, r=8.9, n_r=8)
+            geometry_2 = sections.rectangular_section(d=20, b=133)
+            compound = geometry_2.align_center(geometry_1).align_top(geometry_1) + geometry_1
```
Agent6-6-6, line 1055: @connorferster, there is no longer a align_top method. Think example needs to be updated 👍

sectionproperties/pre/sections.py
```diff
@@ -293,224 +622,757 @@ def plot_geometry(self, ax=None, pause=True, labels=False, perimeter=False):
...
+        :return: CompoundGeometry object rotated by 'angle' about 'rot_point'
+        :rtype: :class:`sections.pre.sections.CompoundGeometry`
+        
+        The following example rotates a 200UB25 section clockwise by 30 degrees::
+
+            import sectionproperties.pre.sections as sections
+
+            geometry_1 = sections.i_section(d=203, b=133, t_f=7.8, t_w=5.8, r=8.9, n_r=8)
+            geometry_2 = sections.rectangular_section(d=20, b=133)
+            compound = geometry_2.align_center(geometry_1).align_top(geometry_1) + geometry_1
```
Agent6-6-6, line 1026: @connorferster, update align_top as no longer exists in code

sectionproperties/pre/sections.py
```diff
@@ -293,224 +622,757 @@ def plot_geometry(self, ax=None, pause=True, labels=False, perimeter=False):
...
+        :return: Geometry object translated to new alignment
+        :rtype: :class:`sections.pre.sections.Geometry`
+
+        The following example erodes a 200PFC section by 3::
+
+            import sectionproperties.pre.sections as sections
+
+            geometry_1 = sections.i_section(d=203, b=133, t_f=7.8, t_w=5.8, r=8.9, n_r=8)
+            geometry_2 = sections.rectangular_section(d=20, b=133)
+            compound = geometry_2.align_center(geometry_1).align_top(geometry_1) + geometry_1
```
Agent6-6-6, line 1136: @connorferster, update align_top as no longer exists in code

sectionproperties/pre/sections.py
```diff
@@ -293,224 +622,757 @@ def plot_geometry(self, ax=None, pause=True, labels=False, perimeter=False):
...
+        top_geoms_acc = []
+        bottom_geoms_acc = []
+        for geom in self.geoms:
+            top_geoms, bottom_geoms = geom.split_section(point_i, point_j, vector)
+            top_geoms_acc += top_geoms
+            bottom_geoms_acc += bottom_geoms
+        return (top_geoms_acc, bottom_geoms_acc)
+
+
+    def offset_section_perimeter(self, amount: float = 0, resolution: float = 12):
```
Agent6-6-6, line 1114: @connorferster, should this also be renamed to `offset_perimeter` like in Geometry class to be consistent, as every other geometry manipulation is named consistently between the two geometry classes??

(and all example updated if changed?)

sectionproperties/pre/sections.py
Agent6-6-6: Hi @connorferster
I'm not so sure this is working correctly or as intended with hollow sections, a negative buffer seems to reduce the interior and exterior, and a positive buffer seems to increase both the interior and exterior?
For example
```
geometry1 = sections.circular_hollow_section(200, 10, 40)
geometry1 = geometry1.offset_perimeter(amount=-4)
geometry1.plot_geometry()
```
The original intent was that it reduced the exterior only to model corrosion of a perimeter?

Perhaps another argument is required to distinguish between wanting to increase or decrease interior/exterior/both so as not to limit it's use (for example may want interior to corrode if you're dealing with a pipeline or something)?

sectionproperties/pre/sections.py
```diff
@@ -1,259 +1,587 @@
...
+        print(f"object at: {hex(id(self))}")
+        print(f"Material: {self.material.name}")
+        return self.geom._repr_svg_()
+
+    @staticmethod
+    def from_points(
+        points: List[List[float]],
+        facets: Optional[List[List[int]]] = None,
+        holes: Optional[List[List[float]]] = None,
+        control_points: Optional[List[List[float]]] = None,
```
Agent6-6-6, line 73: @connorferster, curious if there is ever a use case envisaged where you would want to specify by hand a control_point vs automatically generating it to be within the geometry/section. Potentially seems a little redundant, and one of the benefits to going to shapely was to have them autogenerated vs having to think about specifying them manually?

sectionproperties/pre/sections.py
```diff
@@ -1,259 +1,587 @@
...
+        self.compile_geometry()
+
+    def _repr_svg_(self):
+        print("sectionproperties.pre.sections.Geometry")
+        print(f"object at: {hex(id(self))}")
+        print(f"Material: {self.material.name}")
+        return self.geom._repr_svg_()
+
+    @staticmethod
+    def from_points(
```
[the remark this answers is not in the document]
Agent6-6-6: Actually, I see now from some of your examples that adding material is achieved by `geometry.Material = concrete` for example to apply a material after initially defining a geometry. 

But the main point remains regarding adding material to the add_points as an argument so all the in-built section methods are consistent 👍

sectionproperties/pre/sections.py
connorferster: Hi @Agent6-6-6,

Thanks for checking this out. I confirm that I am seeing the same thing. That is correct, `.material` is not implemented on CompoundGeometry. The intention is that individual geometries would be assigned materials and then combined to create a compound. 

The problem was that I had changed the name of the method from `.offset_section_perimeter()` to `.offset_perimeter()` but I forgot to change the method name in `CompoundGeometry`, also. Now, the method runs. 

You bring up a good point about the method. Agreed, it would be better if you could specify whether to offset the exterior, the interiors, or both. I am working on implementing this now. Will be coming in next commit.

sectionproperties/pre/sections.py
connorferster: Got 'em both now.

sectionproperties/pre/sections.py
connorferster: And that one!

sectionproperties/pre/sections.py
connorferster: Agreed! Updated and fixed.

sectionproperties/pre/sections.py
connorferster: Will create a test for it also.

Agent6-6-6: Ah, I see. 

See one of my last replies in #21 as I think there is a way to automatically create the control points within each bit of geometry. In the case of your composite column case, recognising that the concrete part of the section actually contains a hole in reality where the steel is, you can split up/process the section into the constituent bits without any overlaps and then assign the control points automatically. 

If you could get something like that working it almost means the control_point concept is just an internal requirement that is worked out to aid in the meshing process. Then the user doesn't need to worry about it at all (ideal world) .... I'm sure I'm over simplifying things though!!

connorferster:
Hi @Agent6-6-6, 

I have updated the code to include the ability to assign a control point. For now I think having assignable control point that "travels" with the section allows for the greatest flexibility. I hear what you are saying in regards to a composite column (with the steel actually filling a void within a concrete column). However, that will be up to the user to decide as they build the geometry: I don't think we could come up with a rule that would work in all situations without making things complicated. The new preprocessor allows for the creation of invalid or ambiguous geometry which the designer is required to make unambiguous. 

So the user would have to do something like:
```python
(conc_section.assign_control_point([x,y]) - i_section) + i_section
```

sectionproperties/post/post.py
```diff
@@ -18,18 +19,21 @@ def setup_plot(ax, pause):
...
-def finish_plot(ax, pause, title=''):
+def finish_plot(ax, pause, title="", size=1000, dpi=96):
     """Executes code required to finish a matplotlib figure.
 
     :param ax: Axes object on which to plot
     :type ax: :class:`matplotlib.axes.Axes`
     :param bool pause: If set to true, the figure pauses the script until the window is closed. If
         set to false, the script continues immediately after the window is rendered.
-    :param string title: Plot title
+    :param str title: Plot title
```
Czarified, line 29: You added `size` and `dpi` into the default arguments. Should they be added to the docstring as well? It's obvious for anyone who's worked with matplotlib, but the same could be said for the `title` arg...

```python
:param int size: Plot size
:param int dpi: Plot dpi
```

sectionproperties/pre/dxf.py
```diff
@@ -0,0 +1,34 @@
...
+        import cad_to_shapely as c2s  # type: ignore
+    except ImportError as e:
+        print(e.message)
+        print("To use 'from_dxf(...)' you need to 'pip install cad_to_shapely'")
+        return
+
+    if not dxf_filepath.exists():
+        raise ValueError(f"The filepath does not exist: {dxf_filepath}")
+
+    # TODO avoid step of making a temp file locally
```
Czarified, line 22: Doesn't look like we're writing a temp file now. Does this function actually save a temp file, or is this comment outdated?

Czarified (commented):
I haven't reviewed the examples, or tests, and I only skimmed the `sections.py`, which is actually the biggest change here. I just don't have enough time right now to really dig in. From the past months where I've played with examples and read your comments/thoughts, I think the UX with this change will be significantly improved.

Requiring `cad_2_shapely` for dxf makes sense, but for whatever reason it doesn't sit well with me. Should we include it as an actual requirement, instead of a custom error message? It's a rather small library by our friend @aegis1980, so it may just make sense to include as part of sectionproperties >2.1, if they desire to merge it in.

I'll try to review the rest of it this week, and play around with some examples as well. No promises. 😄 I love all the changes you've made, and really want to see v2.0.0 this year!

sectionproperties/post/post.py
connorferster: 100% Agree. Thanks for the catch!

sectionproperties/pre/dxf.py
connorferster: Yeah, I was not sure what's going on there. It does not seem like we are writing a temp file so I can remove that.

connorferster:
@Czarified I agree with your thoughts on `cad_2_shapely`. I think it would be good to include it as an actual dependency and a full-fledged feature of sectionproperties instead of as a kind of optional add-on. @aegis1980: do you have any thoughts/feelings/preferences on either making your library a dependency of sectionproperties or having sectionproperties vendor it in directly? It works wonderfully and is a great feature to have as part of sectionproperties. Are you planning on developing it further and maintaining it?

Agent6-6-6:
> Hi @Agent6-6-6,
> 
> I have updated the code to include the ability to assign a control point. For now I think having assignable control point that "travels" with the section allows for the greatest flexibility. I hear what you are saying in regards to a composite column (with the steel actually filling a void within a concrete column). However, that will be up to the user to decide as they build the geometry: I don't think we could come up with a rule that would work in all situations without making things complicated. The new preprocessor allows for the creation of invalid or ambiguous geometry which the designer is required to make unambiguous.
> 
> So the user would have to do something like:
> 
> ```python
> (conc_section.assign_control_point([x,y]) - i_section) + i_section
> ```
> 
> As far as this PR goes, I believe it is largely complete (barring any other bug fixes, of course!). I am focusing now on continuing to update all of the documentation to reflect all the new features. I am also planning a new section called "Advanced Geometry Manipulation" to describe some of the issues and techniques that we have been talking about. Things like how to use shape operations to create intermediate nodes.

Yeah I get what you're saying, perhaps a feature for the future. I still believe there is a way to achieve recognising each overlapping geometry and adding control points intdrnally/automatically, I just haven't had much luck with getting it to work as intended, apart from with two geometries as I indicated previously. I need to give it a bit more of a serious effort sometime in the future.

I wonder in the interim if something to report back to the user regarding the meshing not working might consist of reporting the max mesh size actually formed during meshing, vs the defined mesh size. This would highlight that the user that the defined mesh of say 5mm^2 max size wasn't achieved by reporting back that it was 25mm^2 for example. This may indirectly indicating a potential meshing issue involving control points not working due to bits of resulting overlapping section not having or ending up with a control point within them? This might prompt the user to review the geometry/mesh plots if they were not already and manually add further control points as required to mesh things correctly. 

I'd just prefer to see something warning user one way or the other that the mesh went a bit off the rails. At the moment it relies on the user recognising this by reviewing plots which is a bit archaic. To streamline things after initially setting up a script one might disable the plotting for example.

aegis1980:
Hey all, 
Sorry. I turned off notification etc for `section-properties`, but just got a notification about `cad-to-shapely` and found this discussion. 
I have no plans for working on `cad-to-shapely` really. Will just let it (and repo) sit there. Was actually developed to aid another unrelated project and I thought I could use it with `section-properties` (and ran into the meshing issues and gave up).
My inclination would be to keep it as an independent dependency since perhaps Mr/Mrs SomeOne might find a use for cad_to_shapely and develop which could then benefit section-properties and whatever else.
@connorferster As an aside, did you try running the example dxfs (tricky aluminium extrusions) in cad-to-shapely repo through updated section-properties?

connorferster:
Closing PR to merge instead onto v2 branch

PR #78, opened 2021-09-13
Added principal major and minor stress outputs & docs

ccaprani:
I think this can be PR-ed without conflicting with the v2 plans.


Czarified (commented):
I agree. These changes do not seem to conflict with the ongoing v2 work. Most of the changes here are just the black formatting. The two new methods for StressPost build on top of the pre-existing methods 👍🏻 . You could even leave the black formatting, if you wanted. I think we're moving that direction in the future.

ccaprani:
Cheers @Czarified . I undid the black formatting after an errant save autoformat! I guess once the style is agreed for v2 we'll just pipe everything through a CI with style formatting in any case. The big battle will be agreeing line length 🤣

PR #86, opened 2021-09-22
Shapely preprocessor

connorferster:
To merge into v2 branch


connorferster:
Having too much trouble with this...

PR #92, opened 2021-09-24
Docs/index installation changes

robbievanleeuwen:


connorferster:
That all makes good sense! Thanks!

PR #93, opened 2021-09-24
Structure of Analysis Docs

robbievanleeuwen:
Line 33: `create_mesh` now returns a `Geometry` object I think? Worth double checkin!


connorferster (approved):
Yes, it returns a Geometry now which allows for a "fluent API" (I.e. "method chaining"). It's pretty fun to use!

connorferster:
I have been just merging these requests because that's what I normally do at work when I receive a review request but it occurred to me that you might want to merge it yourself! Let me know if you would prefer to be the one doing the merging or if it's okay for me to do so on the review. Happy either way!

robbievanleeuwen:
Happy for you to merge if you don't see anything wrong with it :)

Issue #101, opened 2021-09-26
Compound geometry rotation


normanrichardson:
**Describe the bug**

There is an issue with [`CompoundGeometry.rotate_section`](https://github.com/robbievanleeuwen/section-properties/blob/b775277c4f438ff2b4def30c27115c37d36731d5/sectionproperties/pre/sections.py#L1155-L1182) where under the default parameters (rot_point = 'center') the individual geometries rotate about their own centers.  

**To Reproduce**

```python
import sectionproperties.pre.sections as sections
import matplotlib.pyplot as plt

a = 1
b = 2
t = 0.1

points = [[-t/2, -2*a], [t/2, -2*a], [t/2, -t/2], [a, -t/2], [a, t/2],
          [-t/2, t/2], [-b/2, -2*a], [b/2, -2*a], [b/2, -2*a-t],
          [-b/2, -2*a-t]]
facets = [[0, 1], [1, 2], [2, 3], [3, 4], [4, 5], [5, 0], [6, 7], [7, 8],
          [8, 9], [9, 6]]
holes = []
control_points = [[0, 0], [0, -2*a-t/2]]

geometry = sections.CompoundGeometry.from_points(points, facets, holes, control_points)

fig, (ax1,ax2) = plt.subplots(2)
geometry.plot_geometry(ax=ax1)
geometry = geometry.rotate_section(angle=30, rot_point='center')
geometry.plot_geometry(ax=ax2)
plt.show()
```
<div align='center'>

![issue](https://user-images.githubusercontent.com/14271499/134794843-e24e9a0c-7054-4f34-a9cd-6282dcc96a34.png)

</div>

**Expected behavior**

That the section rotates about the compound geometries center.

robbievanleeuwen:
I would also expect the section to not skew upon rotation? i.e. angle between two adjacent facets to remain 90deg.

I would expect the rotation behaviour to mimic [this ](https://sectionproperties.readthedocs.io/en/latest/rst/examples.html#mirroring-and-rotating-geometry)example.

normanrichardson:
Yeah exactly, I was having issues with the new version of that example. 

I don't think the rotate operation is skewing the geometry, though I have not checked this... Essentially it is calling this: https://shapely.readthedocs.io/en/stable/manual.html#shapely.affinity.rotate

I think the skewing that is showing up in the image is just my terrible plotting :smiley:.

robbievanleeuwen:
[...]

PR #102, opened 2021-09-26
Fix: Compound Rotation Operation

normanrichardson:
Fix applies only when rot_point='center'. Generate the Multipolygons bounding box, and find its centroid. Then proceed to rotate each polygon around this point.

Resolves: #101


connorferster (approved):
Lol, before seeing this PR I wrote the exact same lines of code ;)

Thanks for doing that

PR #108, opened 2021-10-06
Feature: Rhino Importer

normanrichardson:
As initially proposed in #73, this feature aims to:
* Import rhino geometry from `.3dm` files
* Import rhino BREP encodings

Class methods have been added to `sectionproperties.pre.sections.Geometry` and `sectionproperties.pre.sections.CompoundGeometry`. In addition, the Shapely geometry can be accessed via the functions in `sectionproperties/pre/rhino.py`


normanrichardson:
@robbievanleeuwen and @Spectre5 there seems to be a dependency issue that is failing in the CI process. Locally I can resolve this using pip to install the setup.py file. So not using [`python setup.py install`](https://github.com/robbievanleeuwen/section-properties/blob/c2e79668eac48c5710f4317f199a13382f282535/.travis.yml#L11) but rather `pip install .`. I would imagine this is due to some [newer and improved dependency resolution](https://stackoverflow.com/questions/62677190/how-does-pip-determine-the-version-of-a-dependency-of-multiple-dependents/62872812#62872812), but I'm by no means well versed in this and the potential pitfalls. Are you happy if this process changes?

normanrichardson:
@robbievanleeuwen and @Spectre5 thanks for this. I will have a look at #61. 

In `rhino_shapely_interop` I've tried to use [compatible releases](https://www.python.org/dev/peps/pep-0440/#compatible-release) (but perhaps poorly!). My understanding is that these are not as "hard" as pinning, but I think I may need to still reduce those requirements to major releases or perhaps get rid of it altogether.

What is interesting is that `pip install .` is able to resolve these, and `python setup.py install` finds this restrictive.

normanrichardson:
@connorferster I'm happy it is well received

Issue #112, opened 2021-10-21
CompoundGeometry meshing failing


normanrichardson:
**Describe the bug**
Under certain conditions, the `CompoundGeometry` mesh creation step (`geometry.create_mesh`) is failing. 

**To Reproduce**
```python
import sectionproperties.pre.sections as sections
from sectionproperties.analysis.cross_section import Section

s1 = sections.rectangular_section(1,1)
s2 = sections.rectangular_section(0.5,0.5).shift_section(1,0.25)

geometry = sections.CompoundGeometry([s1,s2])
geometry = geometry.rotate_section(angle=30, rot_point='center')

geometry.plot_geometry()
geometry.create_mesh(mesh_sizes=[0.2, 0.1])
section = Section(geometry)
section.display_mesh_info()
```
[triangle.build](https://github.com/robbievanleeuwen/section-properties/blob/b661bb093497d819bcbdf0e8be3028872c202df3/sectionproperties/pre/pre.py#L115) produces errors in `geometry.create_mesh(mesh_sizes=[0.2, 0.1])`: 
```
Internal error in segmentintersection():
  Topological inconsistency after splitting a segment.
  Please report this bug to jrs@cs.berkeley.edu
  Include the message above, your input data set, and the exact
    command line you used to run Triangle.
```

In more extreme cases such as [example_mirr_rot.py](https://github.com/robbievanleeuwen/section-properties/blob/b661bb093497d819bcbdf0e8be3028872c202df3/sectionproperties/examples/example_mirr_rot.py) (after updating to run with v2) the process crashes.

Interestingly if the rotation step is either 0, 90, 180, ... the meshing is successful. For example,
```python
import sectionproperties.pre.sections as sections
from sectionproperties.analysis.cross_section import Section

s1 = sections.rectangular_section(1,1)
s2 = sections.rectangular_section(0.5,0.5).shift_section(1,0.25)

geometry = sections.CompoundGeometry([s1,s2])
geometry = geometry.rotate_section(angle=90, rot_point='center')

geometry.plot_geometry()
geometry.create_mesh(mesh_sizes=[0.2, 0.1])
section = Section(geometry)
section.display_mesh_info()
```
```
Mesh Statistics:
--48 nodes
--17 elements
--2 regions
```

**Expected behaviour**
Meshing step completing.

**Additional context**
After some investigation, I think it is because the triangle is having a hard time with the facets:
```python
import sectionproperties.pre.sections as sections
[...]

PR #116, opened 2021-11-15
Adv geom docs update

connorferster:
Update the "Advanced Geometry Creation" page in the docs to address issues brought up in #112


connorferster:
Built docs locally to confirm they are correct.

robbievanleeuwen:
@connorferster v2 readthedocs builds should be working again with 48b6989.

PR #119, opened 2021-11-19
Only calculate shape factor if material properties have not been defined

robbievanleeuwen:


connorferster:
Would you be willing to write a unit test to describe the bug? Kind of like a test that would fail right now but is fixed with this PR?

PR #120, opened 2021-11-19
Print J & Iw as E.J & E.Iw in display_results() for consistency

### LINKED ISSUE #76: Torsion constant when using materials
More of an FYI in advance of v2.0: when using two materials in a cross section, `display_results` shows the torsion constant as 
> G.J = xxx

the reported quantity is (in fact) E * J, which is consistent with the elastic modulus weighting of other composite cross section properties, but the use of "GJ" may be misleading. So the torsion constant should be found like `section.get_j()/precast.elastic_modulus` if the `precast` concrete is your reference material.

More broadly, since the `Material` class is not 'smart' a little hack is to just use `E = 1` for your reference material and `E = modular_ratio` for the secondary material (the Poisson's ratios should still be correct though for shear-related properties).

robbievanleeuwen:
Closes #76.

### DIFF
--- a/sectionproperties/post/post.py
+++ b/sectionproperties/post/post.py
@@ -190,22 +190,14 @@ def print_results(cross_section, fmt):
     j = cross_section.get_j()
     if j is not None:
         if list(set(cross_section.materials)) != [DEFAULT_MATERIAL]:
-            print(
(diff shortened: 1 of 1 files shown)


robbievanleeuwen:
Assumptions to be outlined in 'Viewing the Results' docs to come.

Issue #88, opened 2021-09-23
v2 sectionproperties logo


robbievanleeuwen:
New v2 logo in sectionproperties has three red dots and white space above, see below screenshot:

<img width="775" alt="Screen Shot 2021-09-23 at 12 44 27 pm" src="https://user-images.githubusercontent.com/18841910/134447719-752fad9c-de3f-4ed3-be55-287910e6e7d1.png">

I'm guessing the white space was added for appropriate github repo logo requirements(?), but looks a bit funny on the home page of the docs? Maybe a different image file could be used for the docs homepage?

I know, really focussing on the big issues...

robbievanleeuwen:
Further, there is transparency above the logo which doesn't work so well with backgrounds that aren't white (e.g. default rtd theme or if dark mode is activated).

<img width="882" alt="Screen Shot 2021-09-23 at 12 49 35 pm" src="https://user-images.githubusercontent.com/18841910/134448084-120967fb-e7f0-425a-a0bf-0a1a210f790a.png">

connorferster:
Ack, I obviously did a poor job of generating the PNG: I actually fussed pretty hard on trying to get it of decent quality but I don’t think I was doing it properly.

connorferster:
Will try again and let you know if I find it getting broken somewhere in the process

robbievanleeuwen:
Fixed with #135

PR #135, opened 2021-12-23
Fix sectionproperties logo

robbievanleeuwen:
Fix for #88.


connorferster:
Hi Robbie, 

I feel silly but the reason I have not completed this task (as originally assigned to me) is because I do not know how I did it in the first place. All efforts I made to create a new logo (again), which involved trying to customize the matplotlib plot remove axes and titles, etc. I could not do so because the stress plots do not accept an axis as an optional argument. For whatever reason, Jupyter would not allow me to modify the axes generated from the stress plot methods. I have spent many hours on this but with nothing to show for it!

How did you go about doing it?

When I created the first stab at a revised logo, I used another stress condition and colour map, however, I have since thought it would be good to recreate the original logo but with the updated warm/cool colour map. What are your thoughts on this?

robbievanleeuwen:
Oh no! I think originally I put a bit of torsion on a box section, got the stress and used a bit of photoshop to make the logo, nothing fancy really. With this edit I just used photoshop again to make a few small changes. If you want to use a different colour scheme happy for you to just send me a script or png of the entire plot and I can do some quick image editing to make it into a logo?

connorferster:
The only reason that I would want to change the colour map for the logo is for consistency with the new default. Aesthetically, I think this other colour map looks better, though. And a logo should be aesthetic! :)

Thanks for doing that. I really appreciate it! I have No. Idea. where those red dots came from. It was so weird.

Issue #126, opened 2021-12-16
Plot figure size not adjusting well


robbievanleeuwen:
**Describe the bug**
In v1, the figure size adjusted quite well to the plot contents. Plotting figures now either cuts off numbers/labels or is too large.

**To Reproduce**
Centroid & stress plots from examples under 'Running an Analysis' docs.

**Expected behaviour**
In v1 plot figure size was adjusted to fit the size of plot with minimal white space and no clipping.

**Screenshots**

<img src="https://user-images.githubusercontent.com/18841910/146296149-b9fb3fbe-c19a-438a-8be8-80f92f917fea.png" width="600" />
<img src="https://user-images.githubusercontent.com/18841910/146296155-38d77a97-840e-4aec-aee9-8a1cf1de27e0.png" width="600" />
<img src="https://user-images.githubusercontent.com/18841910/146296159-bf849ce5-81f6-49bc-a27a-24e95f99c22c.png" width="600" />

**Desktop (please complete the following information):**
 - OS: mac
 - python: 3.8

connorferster:
The plots are something that I would like to fine tune, too. By adding the ability to change the size, the user also has to select an appropriate dpi and fiddling with my two "knobs" makes it harder. I am certainly open to suggestions for improvement!

robbievanleeuwen:
@connorferster quick one - was there a reason you removed the legend for `plot_geometry()` that displayed points/facets, control points & holes?

connorferster:
Probably back in the day. But it would have been a bad reason. I was wondering yesterday why I removed it!

robbievanleeuwen:
Fixed in #138.

PR #138, opened 2021-12-24
Modify figure size behaviour

robbievanleeuwen:
Modifies the default figure size behaviour.

New behaviour changes figure size & dpi after matplotlib's scaling methods, `ax.set_aspect("equal")` & `plt.tight_layout()`, are called. Fixes cropping of axes & colorbars described in #126.

Smallest figure dimension is set to `size` and other dimension set to maintain the ratio determined by matplotlib.

Not convinced this is 100% there as for some geometry aspect ratios there is some extra white space which I believe is caused by `ax.set_aspect("equal")` & `plt.tight_layout()` not working as well as it should together but couldn't figure out a better method. The ideal behaviour would be cropping the white space which `.savefig(filename, bbox_inches='tight')` acheives but I couldn't figure out how to replicate this in interactive plotting mode. Overall I think this is at least an improvement on the current implementation.

Open to thoughts/suggestions on this.


connorferster:
Just started a branch myself to work on the text sizes. Will take a look at this and integrate your updates!

robbievanleeuwen:
Looks good :+1:

connorferster (approved):

PR #139, opened 2021-12-25
.from_points() minor edits

robbievanleeuwen:
- holes parameter not behaving as optional without default assignment.
- re-ordered docs to align with variables
- consistent default material behaviour (?)

### DIFF
--- a/sectionproperties/pre/sections.py
+++ b/sectionproperties/pre/sections.py
@@ -107,7 +107,7 @@ class Geometry:
         points: List[List[float]],
         facets: List[List[int]],
         control_points: List[List[float]],
-        holes: Optional[List[List[float]]],
+        holes: Optional[List[List[float]]] = None,
         material: Optional[pre.Material] = pre.DEFAULT_MATERIAL,
     ):
         """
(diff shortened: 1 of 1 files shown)


sectionproperties/pre/sections.py
```diff
@@ -107,7 +107,7 @@ def from_points(
         points: List[List[float]],
         facets: List[List[int]],
         control_points: List[List[float]],
-        holes: Optional[List[List[float]]],
+        holes: Optional[List[List[float]]] = None,
```
connorferster, line 110: Thanks for the catch!

sectionproperties/pre/sections.py
```diff
@@ -1098,11 +1102,11 @@ def from_points(
         facets: List[List[int]],
         control_points: List[List[float]],
         holes: Optional[List[List[float]]] = None,
-        materials: Optional[List[pre.Material]] = None,
+        materials: Optional[List[pre.Material]] = pre.DEFAULT_MATERIAL,
```
connorferster, line 1105: Agreed. Better to show it this way.

connorferster (approved):

PR #140, opened 2021-12-25
Perimeter & offset tests; CompundGeometry offset fix

robbievanleeuwen:
Summary of changes:
- Added tests for perimeter calculation for `Geometry` and `CompundGeometry`
- Desired behaviour for `CompoundGeometry` offset is for only external perimeter to erode. Current behaviour is that each individual Geometry is offset. Suggested modification to code to achieve this, see `CompundGeometry.offset_perimeter()` and `test_offset.py`.

@connorferster, keen to get your thoughts to see if this method would work? An issue may be that `CompundGeometry.offset_perimeter()` would now return a `Geometry` object, not sure if you see that as an issue?


sectionproperties/pre/sections.py
```diff
@@ -826,13 +824,7 @@ def plot_geometry(self, ax: Optional[matplotlib.axes.Axes]=None, labels=["contro
...
 
         for (i, f) in enumerate(self.facets):
-            if perimeter:
-                if i in self.perimeter:
-                    linewidth = 0.3*size/dpi
-                else:
-                    linewidth = 0.15*size/dpi
-            else:
-                linewidth = 0.15*size/dpi
+            linewidth = 1.5
```
robbievanleeuwen, line 827: Should change this back to `linewidth = 0.15*size/dpi`

sectionproperties/pre/sections.py
```diff
@@ -1499,11 +1491,16 @@ def offset_perimeter(self, amount: float = 0, where = "exterior", resolution: fl
...
-            geoms_acc.append(geom.offset_perimeter(amount, where, resolution))
-        new_geom = CompoundGeometry(geoms_acc)
-        return new_geom
+        # geoms_acc = []
+        # for geom in self.geoms:
+        #     geoms_acc.append(geom.offset_perimeter(amount, where, resolution))
+        # new_geom = CompoundGeometry(geoms_acc)
+        # return new_geom
+
+        unionized_poly = unary_union([geom.geom for geom in self.geoms])
```
robbievanleeuwen, line 1500: Taking a leaf out of the `CompoundGeometry.calculate_perimeter()` book, seems to convert to a single `Polygon` object which we can then use to offset as per the `Geometry` method?

connorferster: I have an idea...

robbievanleeuwen (commented):
Looks good @connorferster! Let me know if you're happy with this implementation and I'll finish off the tests and we can get this one merged as well!

robbievanleeuwen:
Can you share an example of where the ballooning leads to unexpected results? I tried the example you described and the geometry is well depicted. Are you more referring to how material properties are transferred into this new section? If so I wouldn't be too worried about that as I don't think there would ever be a use case for how the 'fusion' of two different materials works. I think the best thing to do would be to alert the user to this and make sure they check their material mesh. 99% of the time this function will be used to examine corrosion of steel.

connorferster:
Hi @robbievanleeuwen! 

Ah, yes, I think I described that incorrectly. You would have to balloon more than the depth of the flange. For some reason, it seems to happen at an offset of 17 in the example below (example shows 20, 17 is the start of ambiguous behaviour, 16 is ok). I very much agree, the primary application would be for corrosion on steel so this is an extreme example (what would that even represent to balloon the material out by this much???)

```python
mat1 = Material("Steel", 200, 0.3, 300, 1000, "red")
mat2 = Material("OtherSteel", 200, 0.3, 400, 1000, "grey")

d = 300
b = 150
tf = 10
tw = 8
r = 12
b_p = 250
t_p = 16
ub = sections.i_section(d=d, b=b, t_f=tf, t_w=tw, r=r, n_r=16, material=mat1)
plate = sections.rectangular_section(b=b_p, d=t_p, material=mat2).align_center(ub).align_to(ub, on="top")
geom = ub + plate
geom = geom.offset_perimeter(amount=20, where='exterior')
geom.create_mesh([100])
section = Section(geom)
section.plot_mesh()
```

image

sectionproperties/pre/sections.py
```diff
@@ -1491,15 +1491,42 @@ def offset_perimeter(self, amount: float = 0, where = "exterior", resolution: fl
...
+            new_geom = CompoundGeometry(geoms_acc)
+            return new_geom
+
+        elif amount > 0: # Ballooning condition
+            # This produces predictable results up to a point.
+            # That point is when the offset is so great it exceeds the thickness
+            # of the material at an interface of two materials.
+            # e.g. A 50 deep plate on top of the top flange of an I-section with a flange depth of 10
+            # When the offset exceeds 10 (the depth of the flange at the intersection), the meshed
+            # material regions will become unpredictable.
```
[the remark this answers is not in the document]
connorferster: Agreed!

robbievanleeuwen:
Awesome! I'll write some docs shortly and then we can merge!

sectionproperties/pre/sections.py
```diff
@@ -1500,7 +1500,10 @@ def offset_perimeter(self, amount: float = 0, where = "exterior", resolution: fl
             for geom in self.geoms:
                 # Use symmetric intersection to find the region of the original
                 # that intersects with the eroded unionized shape
-                geoms_acc.append(geom & offset_geom) 
+                try:
+                    geoms_acc.append(geom & offset_geom)
+                except ValueError:
+                    raise ValueError(f"Try a smaller offset. Section as completely eroded away...")
```
robbievanleeuwen, line 1506: While I think this works, there may be some cases where the user may want to model the complete erosion and continue the analysis e.g. a stiffening plate completely corrodes.

Would it be possible to modify the above to allow this?

connorferster (approved):

sectionproperties/pre/sections.py
connorferster: @robbievanleeuwen To clarify, the eroding of the CompoundGeometry erodes the whole section. It does not allow for the eroding of one part of the compound geometry, like a top plate on some section. That would be pretty nifty and I think could be implemented (ie, erode the geometry with index i) but the ValueError catches, I believe, when any of the geometries erode away. Ah, I think I see what you mean. Will investigate and see what I can come up with. 

Also, do you think it would be worthwhile to add the "index" parameter, to offset just one of the geometries within the compund?

robbievanleeuwen: Agreed that would be a good feature, if possible, but above an beyond what I was thinking. This feature could be useful if certain sides were protected against corrosion while others exposed (a common design scenario).

It seems you might have clued onto what I was thinking about, here's a quick unchecked example which throws an error eventually:

```python
from sectionproperties.analysis.cross_section import Section


uc = sections.i_section(d=400,b=400,t_f=25,t_w=25,r=30,n_r=8)
plate1 = sections.rectangular_section(b=500,d=10).align_center(uc).align_to(uc,"top")
plate2 = sections.rectangular_section(b=500,d=10).align_center(uc).align_to(uc,"bottom")
geom = uc + plate1 + plate2

for i in range(12):
    new_geom = geom.offset_perimeter(-i)
    new_geom.create_mesh([100])
    section = Section(new_geom)
    section.plot_mesh()
    # calculate properties etc...
```

connorferster: The error is generated when the section is eroded to the point it becomes a LineString and LineStrings cannot be handled by sectionproperties as an input geometry which is what is causing the error. It's a special case that needs to be filtered out.

robbievanleeuwen: Do you see anything won't with just discarding the LineStrings once they've completely eroded? E.g. remove the `except` but keep the `try`?

connorferster:
That’s generally the plan. Have been working on it this afternoon. Will commit/push this eve.

connorferster:
@robbievanleeuwen I cannot formally request a review from you since you originated the PR but I would appreciate your eyes on this before the merge.

connorferster:
Tried adding the feature to only offset one of the geometries, and it seemed to work, but I found some meshing problems in some funny instances. I did not feel like rooting out the meshing problems so I left it as is (it can be a future feature if requested by others).

robbievanleeuwen:
@connorferster this is so great thanks for all your work on this one, above and beyond! I think there's still a small issue with completely eroding a `Geometry` object. The line `geoms_acc.append(geom & offset_geom)` creates a `GEOMETRYCOLLECTION EMPTY` shapely object which gets added to the `CompoundGeometry`, this then throws an error when trying to create a `Section object. See the below example and output:

```python
from sectionproperties.analysis.cross_section import Section


uc = sections.i_section(d=400,b=400,t_f=25,t_w=25,r=30,n_r=8)
plate1 = sections.rectangular_section(b=500,d=10).align_center(uc).align_to(uc,"top")
plate2 = sections.rectangular_section(b=500,d=10).align_center(uc).align_to(uc,"bottom")
geom = uc + plate1 + plate2

for i in range(12):
    new_geom = geom.offset_perimeter(-i)
    new_geom.create_mesh([100])

    for geom_i in new_geom.geoms:
        print(geom_i.geom)

    print()
    new_geom.plot_geometry()

    section = Section(new_geom)
    section.plot_mesh()
```
*shapely* objects contained in CompoundGeometry after the top and bottom plates are completely eroded:
Screen Shot 2021-12-28 at 11 18 57 pm

error thrown when trying to create a `Section` object:
Screen Shot 2021-12-28 at 11 19 10 pm

If eroded by 11 mm there are still 3 `Geometry` objects left in the `CompoundGeometry`, two of which are `POLYGON EMPTY` objects, this throws the same error when trying to create a `Section` object.

robbievanleeuwen:
@connorferster added a basic test that if we can get to pass should get us over the line 😄

connorferster:
@robbievanleeuwen Thanks for catching that issue. Took me a while to track down the problem today. Added two lines of code and cleared it up!

Issue #24, opened 2020-05-25
Switch over to triangle


robbievanleeuwen:
Meshpy creates a lot of issues for users - will triangle work?

Spectre5:
It seems that some users are having issues with meshpy (#42, among others).  This seems to be one of the reasons to switch to triangle.  But, since meshpy is using triangle anyway, why do we think that using triangle would actually fix the random crash issue?  Meshpy does indicate that the triangle it includes has some minor changes, but I doubt it makes a minor change that would cause the issue.

Do we have a minimum working example that causes issues for multiple people?  If it crashes for me, I can try to dig a little deeper and see if I can debug it closer.  Perhaps the issue is about some poor input geometry that we can add a check/correction for in sectionproperties.

This said, I'm not opposed to switching to triangle at all.  But I'm not sure that it will magically fix the random crashes that some people have experienced.

@connorferster - maybe you have a minimum working example that regularly crashes?  Ideally one that works with current master, but one that works with your shapely changes is fine too.

Spectre5:
As an example, I see that the mesh is built using the code below.  Maybe the `min_angle` or `mesh_order` used here isn't good for some geometries?  At the least, I think these should probably be keyword arguments so that a user could change them.  Probably just accept `**kwargs` and pass that to the `build` method.

```
mesh = triangle.build(
    mesh, min_angle=30, mesh_order=2, quality_meshing=True,
    attributes=True, volume_constraints=True)
```

connorferster:
Hi all,

Have taken a break from sectionproperties this week. Will try and get back into it this weekend. Have been developing an application at work and it has taken a lot of my brain. 

I will take a look at the geometry and see if how that contributes to the problem. I found it odd because I was getting this result off of the rectangular section test, which I thought would be predictable. Will investigate further and see what I can find out. 

Agreed, if both meshpy and triangle are wrapping Triangle, then I don't know what the difference will be, unless, as @Spectre5 points out, the call to triangle itself may require some tweaking. I think it would be nice if we can make meshpy work but I think I will need to learn more about how Triangle works. 

Part of my ambition for working on a shapely interface is to have the geometry "just work" and not require detailed clean-up prior to meshing. I just had a read of the "Troubleshooting" page on the Triangle site and it seems helpful. I will review the problems I am having in more detail and report back: [http://www.cs.cmu.edu/~quake/triangle.trouble.html](http://www.cs.cmu.edu/~quake/triangle.trouble.html).

[...]

PR #143, opened 2021-12-28
Example and docs update

normanrichardson:
* Bring examples up to date with #133 and #136
* Add rhino-shapley-interop to docs dependencies
* Miscellaneous updates


normanrichardson:
@robbievanleeuwen I was hoping that the requirements file would be a quick fix. I checked locally that the requirements file does work. From the failed build logs it looks like rhino3dm needs CMake to build. I found a possible solution in the readthedocs documentation: [Configuration File V2](https://docs.readthedocs.io/en/stable/config-file/v2.html). I plan to read up on this further.

robbievanleeuwen:
Ok looks good! Let me know if you need me to make any changes to the admin settings :+1:

normanrichardson:
@robbievanleeuwen @connorferster @Spectre5 I'm out of idea on getting the docs to build. To me, it seems the current build error is in Triangle and caused by an environmental variable. Interestingly enough it affects only one example. Any ideas are welcome.

Spectre5:
You might try printing the value of `locale.getlocale(locale.LC_NUMERIC)` before running the docs (maybe in the conf.py file), just to see exactly what it is returning.  I'd expect it to be `(None, None)`, which should be fine to pass to setlocale: `locale.setlocale(locale.LC_NUMERIC, (None, None))`.

Another idea would be to set the locale before the examples run.  Again in the conf.py file, you could do `locale.setlocale(locale.LC_ALL, '')`.

It might be an issue with the configuration of the readthedocs server setup.  I'm not really sure why it only happens in the 1 example though.

Note also that setting the locale is not thread-safe, but I don't think meshpy, sphinx, sphinx-gallery, nor sectionproperties are using threads here.  Further, it isn't really recommended to change the local settings more than once (per Python docs), though I can see why the meshpy team may need to here since they call the C code.

If the above suggestion does not help, I'd suggest opening an issue with the meshpy project and ask them to take a look since the error ultimately comes from the meshpy call to setlocale.  Perhaps they have seen similar issues in the past.

connorferster:
@normanrichardson Does it build if the first example is excluded? If so, what could be in the first, simple, example that causes this problem?

robbievanleeuwen:
If *meshpy* continues to cause too many headaches do you think it might be worth jumping on #24 and implementing this for v2?

It shouldn't be a whole lot of work now that `calculate_plastic_properties()` is not mesh reliant.

normanrichardson:
I tried:
* ignore `01_simple.py` => fails in 02_nastran_section.py
* set `LC_NUMERIC` in conf.py => fails in config.py
* set `LC_ALL` in conf.py => fails in 01_simple.py

Spectre5:
I took a quick look at the build logs...off the top of my head I don't have another great suggestion.  I think it only reports that failure on the first one run (i.e. it reported it on 02_ instead of _01 when you removed the 01_simple script).  The locale `('en_US', 'UTF-8')` is valid, at least on Linux.  Some locales won't work well on Windows that do work on Linux.  But whatever gets reported by getlocale should be fine on the same system.  I wonder if readthedocs does something unusual?

Removing meshpy in favor of triangle is a potential option, but again I see why they do make sure it is a C locale when triangle runs, so we'd maybe want to consider doing it ourselves anyway.  Or else we could just document that unexpected results may occur if you don't ensure it is C locale first (and/or add more tests to ensure other locales work).

One more thing I would try --- is printing the result of getlocale again in the 01_*.py file itself, just to remove the possibility of it changing between conf.py being run and the example python file running.  A new Python process probably does start for running the example files.

robbievanleeuwen:
@normanrichardson we've merged the triangle branch into v2 now with #153. Not sure what the best way to merge those changes into this branch is, but maybe worth having a shot with using triangle over meshpy for building the docs?

connorferster:
> @normanrichardson we've merged the triangle branch into v2 now with #153. Not sure what the best way to merge those changes into this branch is, but maybe worth having a shot with using triangle over meshpy for building the docs?

Yes, you can pull the `v2` changes into your branch and then (hopefully) the docs should build nicely.

connorferster:
I will try merging the v2 branch into @normanrichardson's branch and see what happens...

setup.py
```diff
@@ -4,48 +4,61 @@
...
+    author_email="robbie.vanleeuwen@gmail.com",
+    license="MIT",
     packages=[
-        'sectionproperties', 'sectionproperties.analysis', 'sectionproperties.post',
-        'sectionproperties.pre', 'sectionproperties.tests'
+        "sectionproperties",
+        "sectionproperties.analysis",
+        "sectionproperties.post",
+        "sectionproperties.pre",
+        "sectionproperties.examples",
```
robbievanleeuwen, line 58: Think this needs to be removed now

connorferster:
Nice! The docs builds!

connorferster:
Alright! Now we are cooking with gas! It looks like all tests pass (except the black).

Do you think we are good to merge this @robbievanleeuwen?

robbievanleeuwen:
Awesome! Yeah probably worth just merging straight to v2 instead of docs/examples.

normanrichardson:
Great to see it coming together

Issue #80, opened 2021-09-16
Switch to GitHub Actions


robbievanleeuwen:
### Discussed in https://github.com/robbievanleeuwen/section-properties/discussions/60

<div type='discussions-op-text'>

<sup>Originally posted by **Spectre5** February 12, 2021</sup>
@robbievanleeuwen what are you thoughts on switching from travis CI to the native github actions?  After we add some lint/code quality checks to the repository, I'll follow up with another PR to add those checks to the CI pipeline.  Travis is fine, I'm just more familiar with github actions.  I'm mostly curious if there is a specific reason you're using travis instead?</div>

robbievanleeuwen:
Implemented with #148

PR #148, opened 2021-12-29
Add github actions

robbievanleeuwen:
Begins to address #80. Not diving in too deep yet, have added the following github actions as a starter:

- Lint - lints with black and prepares a PR if there are changes that need to be made
- ~~Test installation - ensures the package can be installed on all operating systems~~
- Run tests - runs all the tests for *sectionproperties* on all operating systems
- Code coverage - generates a codecov report, encouraging further tests to be added


robbievanleeuwen:
@ccaprani @Spectre5 keen to get your input as well if you have any pointers. The idea is to get the ball rolling with v2 and continue to add to the actions in the future as required.

.github/workflows/install-package.yml
```diff
@@ -0,0 +1,27 @@
...
+    - name: Setup python ${{ matrix.python-version }}
+      uses: actions/setup-python@v2
+      with:
+        python-version: ${{ matrix.python-version }}
+
+    - name: Update pip
+      run: pip install --upgrade pip
+
+    - name: Install package
+      run: python -m pip install -e .[test]
```
Spectre5, line 27: Do you need to install the [test] extra here?  If simply making sure sectionproperties itself can install, then it probably isn't needed.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,29 @@
...
+
+on: [workflow_dispatch, push, pull_request]
+
+jobs:
+  tests:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: ['3.8']
```
Spectre5, line 11: I think even better would be to remove the install-package action and instead just run the tests for all 3 OSes (and potentially for multiple python versions like 3.7, 3.8, 3.9, 3.10 --- or maybe just the newest 3, or just the oldest and newest supported ones).  This would obviously make sure it can install on each platform and also make sure that all code runs correctly on all platforms too.  It would make the CI pipeline take longer, but only marginally if they all run in parallel and get started at similar times.

It may require some small modifications to some tests to make sure the result being compared is agnostic to the OS being used, but that is probably best practice anyway.  If any tests are truly unique to an OS, then they can be marked with pytest to only run under certain OSes.

.github/workflows/black.yml
```diff
@@ -0,0 +1,29 @@
...
+
+jobs:
+  build:
+    name: Black with PRs
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Check files using the black formatter
+        uses: rickstaa/action-black@v1
```
Spectre5, line 13: I've never used this github-action, so no big comment her.  However, I've always just installed black and run it myself in the CI environment (along with isort, pylint, flake8, etc, etc).

Spectre5 (commented):
You can see how I approached this in my old PR, which you may have already done.  I had a shell script that could be used to control each of the build tools and the github-actions would try to be minimal and mostly call that script.  The advantage there is that you can use the script locally and thus you should be able to perform the same CI checks locally on your computer too.

Another great project is nox (inspired by tox).  It is an excellent way to run multiple python versions for testing locally (and setup in CI).  And it can have sessions for doing formatting and pretty much any other behavior that you'd like.

.github/workflows/black.yml
robbievanleeuwen: Agreed, was in two minds about this but I think it's best just to use this test to check the code formatting and run the black formatter locally when required.

.github/workflows/tests.yml
robbievanleeuwen: Good idea!

.github/workflows/install-package.yml
robbievanleeuwen: Noted

.github/workflows/black.yml
Spectre5: Ya, this in what I do normally.  You can use --check option for black in CI to check the code only (not format it).  It is probably what this GitHub action does behind the scenes.

Spectre5: So then you can add some logic to format if run locally and check only if running in CI.  Or else explicitly pass the options in each environment.

connorferster (approved):

connorferster:
Does the Github Action script also take care of uploading the new version to PyPI?

robbievanleeuwen:
No haven't added that one yet. Can have a look at that as well today :+1:

ccaprani:
@robbievanleeuwen looks great! An example uploading to pip is [here](https://github.com/MonashSmartStructures/ospgrillage/blob/main/.github/workflows/publish.yml) in case it's of use. More broadly though, you could think about moving the project away from `setup.py` towards the more recent pip `pyproject.toml` [method](https://pip.pypa.io/en/stable/reference/build-system/pyproject-toml/). Things are still in flux moving to this build config method though, so it could be left for a year or two yet either I think.

Issue #22, opened 2020-05-15
meshpy is no longer supported?


BenjaminETaylor:
I'm having trouble installing meshpy on a linux mint machine with Python 3.6.9, and ultimately found [this closed issue](https://github.com/BerkeleyAutomation/meshpy/issues/6) on their github page. In there it says:

> Unfortunately, Python 3+ will not be supported for meshpy. @jeffmahler

Since this conflicts with section-properties' Python 3.5+ requirement, maybe consider switching to [triangle](https://pypi.org/project/triangle/#description) instead? It looks like it uses the same underlying mesh library.

robbievanleeuwen:
Hi Benjamin,

As far as I know meshpy still works fine. Looks like you are looking at the wrong package - [this](https://github.com/inducer/meshpy) is the meshpy that is used by sectionproperties and that is on the python package index.

Robbie

BenjaminETaylor:
Robbie,
Good catch. I did not realize I was looking at a different meshpy! However, I am still not able to install the pypi version of meshpy on my machine. Looks like it's due to a compiler incompatibility (although I don't know why, I have gcc 7.5.0 on my machine). The bigger problem is that this issue also prevents me from installing section-properties!
I looked into it more and I think the [triangle](https://pypi.org/project/triangle/) library does everything section-properties needs, and it installs without issue. I'm willing to give the conversion a shot myself. What are your thoughts? The triangle library has the following benefits:

- fewer end-user installation issues
- lighter weight dependency
- newer releases (latest release -> 04/24/2020)
- same underlying meshing algorithms

robbievanleeuwen:
Hi Benjamin,

Happy to investigate this further as you're not the first person I've come across that has had meshpy installation issues. Everyone I talked to did eventually manage to resolve the issue, but it would be better if no one had problems!

I did initially look at the triangle and the deal breaker was that it didn't generate 2nd order elements (6 node triangles), only 3 nodes. I need these elements for my computation of the section properties so ended up going with meshpy instead. This is a feature of the original triangle written in C so if the developers of the python triangle library included this feature then a switch over to triangle would be more likely!

Robbie

BenjaminETaylor:
Robbie,
[...]

PR #153, opened 2021-12-30
Switch to triangle

robbievanleeuwen:
Swaps out *[meshpy](https://github.com/inducer/meshpy)* for *[triangle](https://github.com/drufat/triangle)* as discussed in #24, #22 and #87.


Spectre5:
Nice!  This will certainly help Windows users since triangle publishes wheels to PyPI and thus they won't have to compile it.

We may want to test or explore any potential issues that using the locale may have, as discussed in #143.

connorferster (approved):

connorferster:
> Nice! This will certainly help Windows users since triangle publishes wheels to PyPI and thus they won't have to compile it.
> 
> We may want to test or explore any potential issues that using the locale may have, as discussed in #143.

I did not see any changes to locale set in the `conf.py` from this branch and it seems that this branch passes all tests and the docs build without special modification. This looks like a winner to me!

Spectre5:
Yes, the locale changes where in meshpy, so it makes sense that we don't have that issue now.  But, if someone writes a program and uses some other locale, when they could, potential, get different results.  Though it would depend on how triangle works, which I haven't gone investigating.

PR #156, opened 2021-12-31
Added bridge sections to library and examples

ccaprani:
@robbievanleeuwen Think I messed up committing to #132, but perhaps you can merge this in and reconcile. I don't think it will work out of the box due to the v2 restructure underway, but hope it is an easy fix for the imports. In a few weeks I can return and tidy up the typehints and get the styles matching better, but thought it might be more helpful to have the skeleton in there now at least.

### DIFF
--- a/sectionproperties/examples/example_bridgebeams.py
+++ b/sectionproperties/examples/example_bridgebeams.py
@@ -1,0 +1,128 @@
+"""
+Examples of using bridgebeams - using Spyder cells
+"""
+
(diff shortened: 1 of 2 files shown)


robbievanleeuwen:
Thanks @ccaprani, I might end up dragging these over to #132 and editing the example once #143 is merged :+1:

ccaprani:
Thanks @robbievanleeuwen . If it's easier just ignore this PR and we'll do it once the library is merged.

I'll also add docs in a few weeks.

PR #166, opened 2022-01-05
Build & deploy changes

robbievanleeuwen:
Changes relating to building the package and auto-deploy:
- Switch from setup.py to pyproject.toml
- Add github action to build and deploy (currently only uploading to TestPyPi)


connorferster:
Will review this week!

connorferster:
I have not learned the github actions yet otherwise I would have made a commit for this: instead of building just on Linux on Python 3.9, would you be up for using the same Python/OS matrix that the tests run on? This way, GH can create pre-built wheels for more setups and make it easier to install on more systems.

Issue #164, opened 2022-01-03
Floating point errors in `.align_to()`


connorferster:
**Describe the bug**
The below code:
```python
import sectionproperties.pre.library.bridge_sections as bridge_sections
import sectionproperties.pre.library.primitive_sections as standard_sections
from sectionproperties.pre.pre import Material
from sectionproperties.analysis.section import Section
from sectionproperties.pre.geometry import CompoundGeometry

Dslab, w, t_f = 180, 2100, 75

precast = Material(
    name="65 MPa",
    elastic_modulus=37.4e3,
    poissons_ratio=0.2,
    yield_strength=65,
    density=2.4e-6,
    color="grey",
)
insitu = Material(
    name="40 MPa",
    elastic_modulus=32.8e3,
    poissons_ratio=0.2,
    yield_strength=40,
    density=2.4e-6,
    color="lightgrey",
)

super_t = bridge_sections.super_t_girder_section(girder_type=5, w=w, material=precast)
slab = standard_sections.rectangular_section(d=Dslab, b=w, material=insitu).align_center(super_t).align_to(super_t, "top")

geom = CompoundGeometry([super_t, slab])
geom.plot_geometry()
geom
``` 
Produces this...

![image](https://user-images.githubusercontent.com/38441489/147904830-6859a105-3219-4cb7-a94b-7bbeb03a1ea3.png)

But it SHOULD produce this...(Note the red svg denoting "invalid" geometry and the hole indicated on the plot)

![image](https://user-images.githubusercontent.com/38441489/147904879-ef171f5c-715b-4400-bd59-574cfb1e7360.png)

**Additional context**
This seems to be a problem with shapely. I opened an [issue with shapely](https://github.com/shapely/shapely/issues/1279) after realizing that one of their examples in their docs does not reproduce as given (another example of the same problem): https://shapely.readthedocs.io/en/stable/manual.html#polygons

```python
exter = [(0, 0), (0, 2), (2, 2), (2, 0), (0, 0)]
inter = [(1, 0), (0, 1), (0.5, 1.5), (1.5, 0.5), (1, 0)]
(Polygon(exter) - Polygon(inter)) # Creates valid MultiPolygon even though two points in common
```

Produces this...
![image](https://user-images.githubusercontent.com/38441489/147905075-3c089fcd-843b-4919-8f96-8c156cd4de54.png)

**Something else that's funny (odd, not 'ha ha')**

[...]

PR #167, opened 2022-01-08
Fix float errors

connorferster:
To address #164, this branch now works on the super_t bridge section example.

Updates:
* Added `geometry.round_polygon_vertices()` function
* Added call to `geometry.round_polygon_vertices()` to `Geometry.__init__()`. This ensures that shapely geometry and `.points` are exactly the same at time of instantiation.

Other updates were made to the library modules to change how `geometry.Geometry`, etc. was imported into the namespace. It seemed that when using IPython autoreload module, there was a namespace clash that would cause an error. By changing the way the module was imported, this clash was fixed.


robbievanleeuwen (approved):

robbievanleeuwen:
Great work @connorferster, thanks for this!

I'm almost done with my plotting PR and then we'll be ready for v2!

PR #168, opened 2022-01-14
Add plotting_context, plotting fixes

### LINKED ISSUE #159: Matplotlib 3.5.x breaks some plots
Matplotlib >= 3.5 seems to break plot sizes and patches in *sectionproperties*. Example below:

```python
import sectionproperties.pre.library.steel_sections as steel_sections

# Create a 150x100x6 RHS on its side
geometry = steel_sections.rectangular_hollow_section(d=100, b=150, t=6, r_out=15, n_r=8)

# Create a mesh and section object. For the mesh, use a maximum area of 2
geometry.create_mesh(mesh_sizes=[2])
section = Section(geometry)

# Perform a geometry and warping analysis

# Perform a stress analysis with Mx = 5 kN.m; Vx = 10 kN and Mzz = 3 kN.m
case1 = section.calculate_stress(Mxx=5e6, Vx=10e3, Mzz=3e6)
```

Plot sizes are much smaller and patches are far less transparent(?) by default (images have been scaled in ratio to their original size - best to view on computer screen but image width and height are both approximately halved):

```python
section.plot_mesh()
```

**matplotlib==3.4.2**

**matplotlib==3.5.1**

```python
case1.plot_stress_vm()
```

**matplotlib==3.4.2**

**matplotlib==3.5.1**

Is the answer to, for now, require matplotlib < 3.5? Patches may be able to be sorted by passing an alpha value as an argument but not sure how to handle the plot size reducing? Haven't had much of a look at it yet thought...

robbievanleeuwen:
This PR fixes #159 and replaces ``setup_plot()`` and ``finish_plot()`` with ``plotting_context()``. ``plotting_context()`` is largely based on the code provided by @Spectre5 in #63, many thanks!

Details below:
- Context manager ``plotting_context()`` - allows saving files, passing kwargs to ``pyplot.subplots()`` amongst other features
- All plotting functions modified to implement ``plotting_context()``, now only return ``ax`` instead of ``(fig, ax)``
- Fixes materials and contours overlaying mesh by adjusting plotting order (i.e. plot contour first, mesh second).
- Removes size & dpi options as these can be passed into **kwargs.
- Adds advanced plotting example to the docs


connorferster (approved):
Geometry.plot_geometry() was missing an "if-clause" to handle labelling holes. Added.

connorferster:
I tried it out and the new plotting works beeeyooootifully. I really like these changes!

PR #189, opened 2022-02-09
Update Util to have 60° radius for bulb

zmpulse:
Update Util to have 60° radius for bulb

### DIFF
--- a/sectionproperties/pre/library/utils.py
+++ b/sectionproperties/pre/library/utils.py
@@ -32,6 +32,7 @@ def draw_radius(pt: list, r: float, theta: float, n: int, ccw: bool = True):
         points.append([x, y])
     return points
 
+
 def draw_radius_60(pt: list, r: float, theta: float, n: int, ccw: bool = True):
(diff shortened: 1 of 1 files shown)


sectionproperties/pre/library/utils.py
```diff
@@ -32,6 +32,35 @@ def draw_radius(pt: list, r: float, theta: float, n, ccw: bool = True):
         points.append([x, y])
     return points
 
+def draw_radius_60(pt: list, r: float, theta: float, n, ccw: bool = True):
```
Czarified, line 35: Why is `n` the only variable with no type hints? Looks like it should always be an int.

`draw_radius_60(pt: list, r: float, theta: float, n: int, ccw: bool = True)`

robbievanleeuwen:
Hi @zmpulse, I incorporated your changes into `draw_radius` as almost all the lines of code for the new method were repeated.

robbievanleeuwen:
Changes merged into #190.

Issue #191, opened 2022-02-13
ModuleNotFoundError: rhino3dm._rhino3dm


connorferster:
**Describe the bug**
In Python 3.9, sectionproperties v2.0.2, there seems to be a bug with the `rhino3dm` module. I have the latest version installed (double checked and all) but it will not import.

I am proposing that we make the rhino importer only import when the `Geometry.from_3dm()` method is called. Currently, I am experiencing an inability to use sectionproperties at all because of this rhino bug :(

**To Reproduce**
Steps to reproduce the behaviour:
1. Install `sectionproperties` in a new Python 3.9.7 virtual environment
2. Run `from sectionproperties.pre.geometry import Geometry`
4. See error, below:

```python
---------------------------------------------------------------------------
ModuleNotFoundError                       Traceback (most recent call last)
~\AppData\Local\Temp/ipykernel_21980/433625605.py in <module>
----> 1 from sectionproperties.pre.geometry import Geometry

~\Anaconda3\lib\site-packages\sectionproperties\pre\geometry.py in <module>
     23 import sectionproperties.pre.bisect_section as bisect
     24 import sectionproperties.post.post as post
---> 25 import sectionproperties.pre.rhino as rhino_importer
     26 
     27 

~\Anaconda3\lib\site-packages\sectionproperties\pre\rhino.py in <module>
      2 from typing import List, Union
      3 from shapely.geometry.polygon import Polygon
----> 4 from rhino_shapely_interop.importers import RhImporter
      5 
      6 

~\Anaconda3\lib\site-packages\rhino_shapely_interop\importers.py in <module>
      4 
      5 import numpy as np
----> 6 import rhino3dm as rh
      7 from shapely.geometry import (
      8     LinearRing,

~\Anaconda3\lib\site-packages\rhino3dm\__init__.py in <module>
      3     from _rhino3dm import *
      4 else:
----> 5     from ._rhino3dm import *
      6 
      7 __version__ = '7.11.0'

ModuleNotFoundError: No module named 'rhino3dm._rhino3dm'
```

**Expected behaviour**
sectionproperties imports

**Desktop (please complete the following information):**
 - OS: Windows 10
 - JupyterLab Desktop

@normanrichardson Have you experienced this at all?

Anyone else seeing this in their environments?

normanrichardson:
[...]

PR #197, opened 2022-02-26
Moved imports for rhino libraries into rhino methods

connorferster:
Moved the rhino imports into rhino related methods so that they are only loaded when rhino import functionality are used.


robbievanleeuwen (commented):
Looks good to me.

Not sure if you meant to add all those notebook files?

connorferster:
Arrrgh!!! I did a git add -A and forgot that I cannot do that on this project.

Thanks for catching that…

connorferster:
Fixes #191

robbievanleeuwen:
Think there's still two additional files trying to make their way into this PR 😆

robbievanleeuwen (approved):

PR #201, opened 2022-02-28
Quick demo of "bar_array" approach for creating RC sections

connorferster:
Created linear_bar_array() function for creating reinforcing in rectangular column sections, rectangular beam sections, or any other type of linear-style concrete section.

The idea is that by creating a CompoundGeometry of the bar array, we can subtract the whole bar array from the concrete portion in one operation.

### DIFF
--- a/sectionproperties/pre/library/concrete_sections.py
+++ b/sectionproperties/pre/library/concrete_sections.py
@@ -1,5 +1,5 @@
 import numpy as np
-from shapely.geometry import Polygon
+from shapely.geometry import Polygon, Point
 import sectionproperties.pre.pre as pre
 import sectionproperties.pre.geometry as geometry
 import sectionproperties.pre.library.primitive_sections as primitive_sections
@@ -9,7 +9,8 @@ def concrete_rectangular_section(
     b: float,
     d: float,
     dia: float,
-    n_bar: int,
+    n_rows: int,
+    n_cols: int,
     n_circle: int,
     cover: float,
     area: float = None,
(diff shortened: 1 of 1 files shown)


connorferster:
Can be used with the following script: 
 
```python
from sectionproperties.pre.pre import Material
from sectionproperties.pre.library import concrete_sections

steel_400 = Material("400 MPa Steel", 200e3, 0.3, 400, 7.7, 'teal')
conc_35 = Material("35 MPa Concrete", 24e3, 0.3, 35, 2.4, 'palegoldenrod')

n_rows = 4
n_cols = 3
col_geom = concrete_sections.concrete_rectangular_section(
    b=400,
    d=600, 
    dia=25, 
    n_rows=n_rows, 
    n_cols=n_cols, 
    n_circle=12, 
    cover=30, 
    conc_mat=conc_35, 
    steel_mat=steel_400
)
col_geom.create_mesh([500] + [100]*(n_rows * n_cols))
col_geom.plot_geometry()
col_sec = Section(col_geom)
col_sec.plot_mesh()
```

robbievanleeuwen:
Nice! Do you think limiting to only one subtraction operation will be an ongoing limitation?

I would rather generalise to allowing a list of bar locations and sizes rather than an array as arrays themselves are quite restrictive.

connorferster:
I wrote out another response just now but, in the writing of it, I got an idea for a (hopefully) more robust difference operation.

Will try implementing it and see how it goes.

Short answer: should be able to do more than one difference op but it may not be the best way to “build up” compound geoms because of the loss of material information.

robbievanleeuwen:
Suggest this gets superseded by #202? Alternatively just keep the `linear_bar_array()` method for users to add their own reinforcement easily?

connorferster:
I will submit a separate PR for the inclusion of linear_bar_array() because it will allow for the design and layout of an arbitrary RC rectangular column section.

PR #202, opened 2022-03-01
Update concrete section definitions

robbievanleeuwen:
New concrete section definitions:
- Now correctly subtracts area for reinforcement, i.e. `geom = (geom - bar) + bar`
- Added option for adding top reinforcement to rectangular and tee sections
- Added option for added circular concrete area for circular section

### DIFF
--- a/sectionproperties/pre/library/concrete_sections.py
+++ b/sectionproperties/pre/library/concrete_sections.py
@@ -8,36 +8,46 @@ import sectionproperties.pre.library.primitive_sections as primitive_sections
 def concrete_rectangular_section(
     b: float,
     d: float,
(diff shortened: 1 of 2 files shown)


robbievanleeuwen:
(seems to be an issue with new version on `rhino3dm` released ~1hr ago?)

connorferster (approved):

connorferster (changes requested):
In the interest of improving test coverage, would you be up for writing a basic test for each the concrete sections?  Something just validating that the geometry expected is what is indeed being produced :)

robbievanleeuwen:
@connorferster just adding some simple area checking tests :)

Issue #216, opened 2022-05-28
Testing the Installation fails for test_sections


simo-11:
**Describe the bug**
Testing the Installation fails for test_sections due to missing files in python package and file referencing errors.
using \_\_file\_\_ instead of pathlib.Path.cwd()/... is one possible solution for referencing files in tests module.

**To Reproduce**
pytest --pyargs sectionproperties

**Expected behaviour**
Test would be run successfully

**Screenshots**
sectionproperties\tests\test_sections.py ......F.FFFF.....                                                                                                             [ 94%]
==================== FAILURES ====================
_____ test_geometry_from_dxf ___
    def test_geometry_from_dxf():
        section_holes_dxf = (
            pathlib.Path.cwd() / "sectionproperties" / "tests" / "section_holes.dxf"

**Desktop (please complete the following information):**
 - OS: Windows-10
 - Version 2.0.3

**Additional context**
section_holes.dxf and other non python files are missing from python package
![image](https://user-images.githubusercontent.com/1210784/170816205-994aa965-bc30-4bb3-b12a-7af602b03c18.png)
![image](https://user-images.githubusercontent.com/1210784/170816225-c4175d2a-4157-4b36-b7e4-f3a2d2b41bb9.png)

robbievanleeuwen:
@simo-11 thanks for raising this! I can reproduce on my end as well.

PR #217, opened 2022-05-29
Fix for #216

simo-11:
Fix replaces use of current working directory by using directory of test file to allow running tests anywhere.
This pull request does not handle packaging issue that leaves data files off.


Czarified (commented):
Nice small fix. We definitely should avoid using cwd where possible. My two cents:

`pathlib` is already imported in the module, and the calls to get the `__file__` directory would be very similar to the `dirname()`.

I believe pathlib is also a more modern standard library module, and generally more preferred. The typical usage is to grab the Path object directly, though, not what we currently have. `from pathlib import Path`

robbievanleeuwen:
@simo-11 thanks for the fix, wondering if the recent two commits are related to this PR?

@Czarified thanks for the input! I'm definitely not up to date with what is more modern. @simo-11 would you be happy to try the fix using ``pathlib``?

simo-11:
Using pathlib is fine. I have just knew dirname and knew it could handle this.

robbievanleeuwen:
@simo-11 thanks. I'm happy to make changes to PR myself but have noticed you have included an example file ``rectangle.py`` in this branch, is there an intention to add this to the library as well?

simo-11:
rectangle.py is my own experiment. I'm not offering currently to be included in main.

robbievanleeuwen:
@simo-11 ok that's fine, are you happy for me to remove ``rectangle.py`` from the master branch in [simo-11/section-properties](https://github.com/simo-11/section-properties)? If not,  I can create a new PR with the changes.

simo-11:
I'm ok with with any solution. Sorry for non clean pull request.

PR #225, opened 2022-06-18
Geometry disjoint fix

connorferster:
I updated the implementation to build a network of all the polygons and then traverse the network to see which ones were not connected. It seemed like the only way 🤷


robbievanleeuwen:
Nice! :eyes: :brain:

Issue #199, opened 2022-02-28
`CompoundGeometry.from_points()` does not assign material properties


robbievanleeuwen:
When creating a compound geometry using the `from_points` technique, material properties are not assigned.

This is because is no material assignment in the `CompoundGeometry.from_points()` method, e.g. see [here](https://github.com/robbievanleeuwen/section-properties/pull/196/files#r815260722).

I have suggested an incomplete fix in #196, but needs additional work for the `else` statement and default value given to the `materials` attribute.

robbievanleeuwen:
Hi @connorferster, I've added some test into a new branch [from_points_material_fix](https://github.com/robbievanleeuwen/section-properties/tree/from_points_material_fix). Looks like the ``Geometry.from_points()`` method works fine, just an issue with the ``CompoundGeometry.from_points()``.

PR #226, opened 2022-06-19
From points material fix

connorferster:
As per #199, updates implementation of `CompoundGeometry.from_points()` to include a list of materials that are assumed to be in the same order as the control_points.


robbievanleeuwen (approved):

robbievanleeuwen:
Awesome! Thanks @connorferster!

Issue #175, opened 2022-01-18
Extract stress at specific location


robbievanleeuwen:
### Discussed in https://github.com/robbievanleeuwen/section-properties/discussions/174

<div type='discussions-op-text'>

<sup>Originally posted by **scote89** January 18, 2022</sup>
Could we please have a function or example code on how to output the stress at a specific coordinate? I do not know how to get position values linked with the stress result to then interpolate.</div>

connorferster:
Could this play off of the method that generated the Mohr's circle? Does that not get the stress at a point?

Czarified:
The original discussion thread was #68. I believe the issue is that the Mohr's Circle function doesn't return values, just the graph. If you want to pull the values and pass them on for anything else, you need to write your own function.

But I think you're right, we could take colin's method and break it up into 2 different methods.

robbievanleeuwen:
There is also a bit of work in perform the correct interpolation based on the Tri6 shape functions. Currently the Mohr's circle method performs a linear interpolation. Once implemented, the Mohr's circle method will call this new get stress method to obtain the stress at a point.

ccaprani:
@robbievanleeuwen I'll give this a go in the next few days. I'll use the shape function interpolations and expose a public API for it, and then update the Mohr's Ciricles accordingly. It would be a good addition indeed!

normanrichardson:
@robbievanleeuwen and @ccaprani I had a look at this problem as I wanted to look at my FEM theory again. I had some issues getting the mapping (for a point in (x,y) to a point in (eta, xi, zeta)) right, but I think it is correct. I think I've made some progress that may be useful: a [method](https://github.com/normanrichardson/section-properties/blob/deca66eea2008d1a7d7a25fb1fb7f5cbba05f3b2/sectionproperties/analysis/fea.py#L558-L708) in `Tri6` that matches `Tri6.element_stress`. As is, it is a bit clunky to use, but it seems to work. 

![Figure_1](https://user-images.githubusercontent.com/14271499/154185597-4e08198e-7e51-4f08-8a44-019e14ac1ee0.png)

To make it easier to use I think some architectural decisions are needed. In particular, how to integrate this into `StressPost`.

Would this be a valuable contribution at this point? 

One technical question I have is if it is correct to use the shape functions to interpolate the element's nodal stresses? As opposed to interpolating the average nodal stresses.

ccaprani:
[...]

PR #230, opened 2022-06-29
Feat stress elm local

normanrichardson:
This is draft for some of basic elements of issue #175. It gives the ability for the Tri6 element to determine its stress at a point within its domain.

Known limitations:
1. The mappings in `Tri6.local_coord` that take a global coordinate and maps it to a local coordinate (or physical coordinate to a natural coordinate) are based on alphine transformations and are not general enough for isoparametric triangular elements. The mappings are sufficient for "straight line" Tri6 elements. I believe the underlying mesher [`triangle`](http://www.cs.cmu.edu/~quake/triangle.html) is limited to "straight line" Tri6's.

In it's current state this can be used as follows:
```python
import sectionproperties.pre.library.steel_sections as steel_sections
import sectionproperties.analysis.fea as fea
import numpy as np
import matplotlib.pyplot as plt
from matplotlib.gridspec import GridSpec

# Create a 150x100x6 RHS on its side
geometry = steel_sections.rectangular_hollow_section(d=100, b=150, t=6, r_out=15, n_r=8)

# Create a mesh and section object. For the mesh, use a maximum area of 2
geometry.create_mesh(mesh_sizes=[100])
section = Section(geometry)
section.plot_mesh()

# Perform a geometry and warping analysis

# Section properties needed for local stress
sect_prop = {
    'ea':  section.section_props.ea,
    'cx':  section.section_props.cx,
    'cy':  section.section_props.cy,
    'ixx':  section.section_props.ixx_c,
    'iyy':  section.section_props.iyy_c,
    'ixy':  section.section_props.ixy_c,
    'i11':  section.section_props.i11_c,
    'i22':  section.section_props.i22_c,
    'phi':  section.section_props.phi,
    'j':  section.section_props.j,
    'Delta_s':  section.section_props.Delta_s,
    'nu':  section.section_props.nu_eff,
}

# Perform a stress analysis with Mx = 5 kN.m; Vx = 10 kN and Mzz = 3 kN.m
state = {
    'N': 0,
    'Mxx': 5e6,
    'Myy': 0,
    'M11': 0,
    'M22': 0,
    'Mzz': 3e6,
    'Vx': 10e3,
    'Vy': 0
}
case1 = section.calculate_stress(**state)

# set up plots
fig = plt.figure()

gs = GridSpec(2, 2, figure=fig)
ax1 = fig.add_subplot(gs[0, :])
ax2 = fig.add_subplot(gs[1, 0])
ax3 = fig.add_subplot(gs[1, 1])

# Plot the bending stress for case1
case1.plot_stress_v_zx(pausr=False, render=False, ax=ax1)

def getElementStess(p):

    # find the Tri6 element containing the point
    tri_id = None
    for (idx , tri) in enumerate(section.elements):
        if tri.point_within_element(p):
            tri_id = idx
    tri = section.elements[tri_id]

    # get the stess at the point
    sigs = tri.local_element_stress(
        p=p,
        **state,
        **sect_prop,
        omega=section.section_props.omega[tri.node_ids],
        psi_shear=section.section_props.psi_shear[tri.node_ids],
        phi_shear=section.section_props.phi_shear[tri.node_ids]
        )

    return sigs

# get and plot the stress at points
s = []
x = []
y = []
for ii in np.arange(5.015,148.99,4):
    # global point of interest
    p = np.array([ii,4.5])
    y.append(p[1])
    x.append(p[0])
    s.append(getElementStess(p)[7])

ax1.plot(x,y, c="green", marker='o', alpha=.6)
ax2.plot(x,s, c="green", marker='o')

s = []
x = []
y = []
for ii in np.arange(5.015,94.99,3):
    # global point of interest
    p = np.array([5, ii])
    y.append(p[1])
    x.append(p[0])
    s.append(getElementStess(p)[7])
ax1.plot(x,y, c="blue", marker='o',alpha=.6)
ax3.plot(y,s, c="blue", marker='o')

plt.show()
```

### DIFF
--- a/sectionproperties/analysis/section.py
+++ b/sectionproperties/analysis/section.py
@@ -27,9 +27,9 @@ import sectionproperties.analysis.fea as fea
 import sectionproperties.analysis.solver as solver
 import sectionproperties.post.post as post
 
-from shapely.geometry import asPoint, Polygon
+from shapely.geometry import Polygon
 from shapely.strtree import STRtree
-
+from shapely.geometry import Point
 
 class Section:
     """Class for structural cross-sections.
@@ -2150,7 +2150,7 @@ class Section:
         Vy=0,
         agg_func=np.average,
     ) -> Tuple[float]:
-        """Calaculates the stress at a point within an element for given design actions
+        """Calculates the stress at a point within an element for given design actions
         and returns *(sigma_zz, tau_xz, tau_yz)*
 
         :param pt: The point. A list of the x and y coordinate
@@ -2168,7 +2168,6 @@ class Section:
             (sigma_zz, tau_xz, tau_yz) are retrieved from each element and combined according to this function.
             By default, `numpy.average` is used.
         :type agg_function: function, optional
-        :return: Resultant normal and shear stresses (sigma_zz, tau_xz, tau_yz)
         :return: Resultant normal and shear stresses list[(sigma_zz, tau_xz, tau_yz)]. If a point it not in the
             section then None is returned.
         :rtype: Union[Tuple[float, float, float], None]
@@ -2191,7 +2190,7 @@ class Section:
         Vy=0,
         agg_func=np.average,
     ) -> List[Tuple]:
-        """Calaculates the stress at a set of points within an element for given design actions
+        """Calculates the stress at a set of points within an element for given design actions
         and returns *(sigma_zz, tau_xz, tau_yz)*
 
         :param pts: The points. A list of several x and y coordinates
@@ -2241,7 +2240,7 @@ class Section:
         }
 
         for pt in pts:
-            query_geom = asPoint(pt)
+            query_geom = Point(pt)
             tri_ids = [
                 self.poly_mesh_idx[id(poly)]
                 for poly in self.mesh_search_tree.query(query_geom)
--- a/sectionproperties/analysis/section.py
+++ b/sectionproperties/analysis/section.py
@@ -30,6 +30,7 @@ import sectionproperties.post.post as post
 from shapely.geometry import asPoint, Polygon
 from shapely.strtree import STRtree
 
+
 class Section:
     """Class for structural cross-sections.
 
@@ -188,7 +189,10 @@ class Section:
         self.mesh_attributes = attributes
 
         # create the search tree
-        p_mesh = [Polygon(self.geometry.mesh["vertices"][tri][0:3]) for tri in self.geometry.mesh["triangles"]]
+        p_mesh = [
+            Polygon(self.geometry.mesh["vertices"][tri][0:3])
+            for tri in self.geometry.mesh["triangles"]
+        ]
         self.poly_mesh_idx = dict((id(poly), i) for i, poly in enumerate(p_mesh))
         self.mesh_search_tree = STRtree(p_mesh)
 
@@ -2134,7 +2138,17 @@ class Section:
         )
 
     def get_stress_at_point(
-        self, pt: List[float], N=0, Mxx=0, Myy=0, M11=0, M22=0, Mzz=0, Vx=0, Vy=0, agg_func=np.average
+        self,
+        pt: List[float],
+        N=0,
+        Mxx=0,
+        Myy=0,
+        M11=0,
+        M22=0,
+        Mzz=0,
+        Vx=0,
+        Vy=0,
+        agg_func=np.average,
     ) -> Tuple[float]:
         """Calaculates the stress at a point within an element for given design actions
         and returns *(sigma_zz, tau_xz, tau_yz)*
@@ -2157,12 +2171,23 @@ class Section:
         :return: Resultant normal and shear stresses (sigma_zz, tau_xz, tau_yz)
         :rtype: tuple(float, float, float)
         """
-        sigs = self.get_stress_at_points([pt], N, Mxx, Myy, M11, M22, Mzz, Vx, Vy, agg_func)
+        sigs = self.get_stress_at_points(
+            [pt], N, Mxx, Myy, M11, M22, Mzz, Vx, Vy, agg_func
+        )
         return next(sigs)
 
-
     def get_stress_at_points(
-        self, pts: List[List[float]], N=0, Mxx=0, Myy=0, M11=0, M22=0, Mzz=0, Vx=0, Vy=0, agg_func=np.average
+        self,
+        pts: List[List[float]],
+        N=0,
+        Mxx=0,
+        Myy=0,
+        M11=0,
+        M22=0,
+        Mzz=0,
+        Vx=0,
+        Vy=0,
+        agg_func=np.average,
     ) -> List[Tuple]:
         """Calaculates the stress at a set of points within an element for given design actions
         and returns *(sigma_zz, tau_xz, tau_yz)*
@@ -2214,10 +2239,14 @@ class Section:
 
         for pt in pts:
             query_geom = asPoint(pt)
-            tri_ids = [self.poly_mesh_idx[id(poly)] for poly in self.mesh_search_tree.query(query_geom) if poly.intersects(query_geom)]
-            if len(tri_ids) ==0:
+            tri_ids = [
+                self.poly_mesh_idx[id(poly)]
+                for poly in self.mesh_search_tree.query(query_geom)
+                if poly.intersects(query_geom)
+            ]
+            if len(tri_ids) == 0:
                 sig = None
-            elif len(tri_ids)==1:
+            elif len(tri_ids) == 1:
                 tri = self.elements[tri_ids[0]]
                 sig = tri.local_element_stress(
                     p=pt,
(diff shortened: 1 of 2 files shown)


robbievanleeuwen:
This looks great @normanrichardson! I think we can get it to work based off what you've done here, and I'm happy with the linear interpolatoin results!

Design-wise, I think we can store the section actions in the ``StressPost`` class and this could be passed to a ``Section.get_stress_at_point(pt, stress_post)`` method. We could also look at the ``stress_post`` (or similarly named) variable being either a ``StressPost`` object or a dictionary of actions if the user doesn't want to run a full stress analysis on the section. I think this is maybe a good idea, but open to your thoughts?

All the section properties could be retrieved from the ``SectionProperties`` class attribute and then we should all have all the relevant information to pass to ``Tri6.local_element_stress()``. I think it makes sense to combine the stresses in  ``Tri6.local_element_stress()`` to only output ``(sigma_zz, tau_xz, tau_yz)`` at previously discussed.

Moving forward, do you have the time/enough information to finish this off? If not, I can have a look into it over the next month or so. Thanks again for all your work on this!

robbievanleeuwen:
Also love the line plots you've done, seems like it should be easy to build a ``plot_stress_along_line`` function or similar!

Czarified:
Thanks so much @normanrichardson ! Great work!

normanrichardson:
@Czarified no worries, thanks for the prodding.

@robbievanleeuwen, yeah I think I can finish this up.

> "store the section actions in the StressPost class"

In `section.calculate_stress` when the [`StressPost`](https://github.com/robbievanleeuwen/section-properties/blob/b9153ae290d63deaba4be71bc782d3c481de4777/sectionproperties/analysis/section.py#L1083) is created I will adjust this to take the in the forces (`N, Vx, Vy, Mxx, Myy, M11, M22, Mzz`) and store these in `StressPost`. Then basically just copy over the function `getElementStess(p)` defined above to `StressPost`. Thoughts beyond may become more clear as I go.

One thought is to change stress_post to calculate and hold the element stresses, and not the averaged nodal stresses, or maybe hold both. But I would hesitate to go that far for now.

> "only output (sigma_zz, tau_xz, tau_yz)"

It has been a while, so I'm a little rusty on the details here. Are these `StressResult.sig_zz`, `StressResult.sig_zx` , and `StressResult.sig_zy`? And can find those combinations [here](https://github.com/robbievanleeuwen/section-properties/blob/b9153ae290d63deaba4be71bc782d3c481de4777/sectionproperties/analysis/section.py#L4719-L4721)?

normanrichardson:
@robbievanleeuwen I think I miss understood. 

Would `Section.get_stress_at_point` be a compliment to `Section.calculate_stress`?

The way I was describing it would work:
```python
# the usual above
state = {
    'N': 0,
    'Mxx': 5e6,
    'Myy': 0,
    'M11': 0,
    'M22': 0,
    'Mzz': 3e6,
    'Vx': 10e3,
    'Vy': 0
}
case1 = section.calculate_stress(**state)
p = np.array([5, ii])
stress_p = case1.get_stress_at_point( p)
```

But what I'm now thinking you're saying:
```python
# the usual above
state = {
    'N': 0,
    'Mxx': 5e6,
    'Myy': 0,
    'M11': 0,
    'M22': 0,
    'Mzz': 3e6,
    'Vx': 10e3,
    'Vy': 0
}
case1 = section.calculate_stress(**state)
stress_p = section.get_stress_at_point( p, **state)
```
Where the type of case1 and stress_p could differ.

Is this closer to what you are thinking?

robbievanleeuwen:
Yes I was thinking the second option because if a user was only interested in the stress at certain points they wouldn't have to run the entire analysis using ``Section.calculate_stress()``. If I understand the implementation correctly, we don't need a ``StressPost`` object at all, just access to the calculated section properties and we are only performing a calculation one one element. I was thinking the ``Section`` class method could look like:

```python
def get_stress_at_point(
    self,
    pt: List[float],
    actions: Dict[str, float],
) -> Tuple[float]:
    """description...

    e.g. pt = [0.5, 1.5]
    e.g actions = state (from your comment above)

    :return: Resultant normal and shear stresses (sigma_zz, tau_xz, tau_yz)
    """

    # do stuff
```

Disregard the comment about saving the actions to the ``StressPost`` object, I think this is a lot cleaner and should be done independently of ``StressPost``.

> It has been a while, so I'm a little rusty on the details here. Are these `StressResult.sig_zz`, `StressResult.sig_zx` , and `StressResult.sig_zy`? And can find those combinations [here](https://github.com/robbievanleeuwen/section-properties/blob/b9153ae290d63deaba4be71bc782d3c481de4777/sectionproperties/analysis/section.py#L4719-L4721)?

Correct, these are just the resultant stresses, calculation as you've highlighted!

normanrichardson:
Potential issue:
I'm not sure I like how points on the element boundaries are handled. Right now it takes the values from the last element it finds, [here](https://github.com/robbievanleeuwen/section-properties/pull/230/commits/a83225047dfae024b958f7b385851e4f3995745d#diff-b4daee97f2caa9384dc65add36e927ff8a6f91520c28a2575c168d3fb9967878R2186-R2190). If I remember correctly, the stress over element boundaries is not continuous. So some sort of averaging may make sense.

Other potential improvements:
The [element search](https://github.com/robbievanleeuwen/section-properties/pull/230/commits/a83225047dfae024b958f7b385851e4f3995745d#diff-b4daee97f2caa9384dc65add36e927ff8a6f91520c28a2575c168d3fb9967878R2186-R2190) is not great. I believe there are ways to speed this up. Initially, I would try using shapely to [generate the mesh polygons](https://shapely.readthedocs.io/en/maint-1.8/manual.html#delaunay-triangulation) and perform [STR-tree search](https://shapely.readthedocs.io/en/maint-1.8/manual.html#str-packed-r-tree). This should help if asking for more than one point (it may also help for just one point). Should I add a function get_stress_at_points that does this?

robbievanleeuwen:
Correct, pretty sure the stresses are not continuous at element boundaries. This is a common theme in FEA with extrapolating stresses from gauss points. If you could average the boundary result based on the two adjacent elements I think that would be suitable?

Regarding the element search, if you think it is worth adding this feature I'm all for it 🙌

Czarified:
From my experience, and referencing a "Practical FEA" textbook, element-nodal bilinear extrapolation with nodally averaged boundary results is common, and with a slight conservatism. I could look up the reference, if you're curious.

normanrichardson:
@robbievanleeuwen let me know your thoughts. Thanks

normanrichardson:
I realized that I need to update the docstring to include the return type of `None`, for the case when a point is outside the section.

robbievanleeuwen (changes requested):
This is great, thanks @normanrichardson! A few small minor spelling/docs items but the rest of it looks good to me!

sectionproperties/analysis/section.py
```diff
@@ -2126,6 +2137,149 @@ def get_sf_p(self):
...
+            "i11": self.section_props.i11_c,
+            "i22": self.section_props.i22_c,
+            "phi": self.section_props.phi,
+            "j": self.section_props.j,
+            "Delta_s": self.section_props.Delta_s,
+            "nu": self.section_props.nu_eff,
+        }
+
+        for pt in pts:
+            query_geom = asPoint(pt)
```
robbievanleeuwen, line 2244: I get the following warning:

> sectionproperties/analysis/section.py:2179: ShapelyDeprecationWarning: The proxy geometries (through the 'asShape()', 'asPoint()' or 'PointAdapter()' constructors) are deprecated and will be removed in Shapely 2.0. Use the 'shape()' function or the standard 'Point()' constructor instead.

robbievanleeuwen: Changing this to ``Point(pt)`` works for me, with ``from shapely.geometry import Point``.

normanrichardson: Got it, I must be running an older version of Shapely I guess

sectionproperties/analysis/section.py
```diff
@@ -2126,6 +2137,149 @@ def get_sf_p(self):
...
+        :type pt: list[float, float]
+        :param float N: Axial force
+        :param float Vx: Shear force acting in the x-direction
+        :param float Vy: Shear force acting in the y-direction
+        :param float Mxx: Bending moment about the centroidal xx-axis
+        :param float Myy: Bending moment about the centroidal yy-axis
+        :param float M11: Bending moment about the centroidal 11-axis
+        :param float M22: Bending moment about the centroidal 22-axis
+        :param float Mzz: Torsion moment about the centroidal zz-axis
+        :param agg_function: A function that aggregates the stresses if the point is shared by several elements.
```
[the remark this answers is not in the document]
normanrichardson: Thought some added flexibility may be nice

Czarified:
Thanks so much, @normanrichardson ! Great work on this feature!

robbievanleeuwen:
Agreed thanks again @normanrichardson! I think this is a really useful addition to *sectionproperties*! I'm hoping to overhaul the docs in the near future and will definitely be adding an example that uses this feature!

normanrichardson:
No worries, any time. It was good to partially dust off the fea knowledge

Issue #188, opened 2022-02-03
Default plot size cutting off wider plot legends


Agent6-6-6:
@robbievanleeuwen this is potentially related to #159. But I've noticed with what appears to be a reduced default size of the plots from matplotlib that some portion of the legends on certain plots are disappearing off the edge of the default plot size. e.g.:-

![image](https://user-images.githubusercontent.com/36808164/152296206-0d062247-2fec-4296-8ce1-77d44a0ce83a.png)

Wondering if a default minimum plot size should be enforced by sectionproperties (or potentially some review and reformatting of the legends) to avoid this occurring in common cases with the default plot size (I could have sworn they were larger by default some time ago when I last used sectionproperties, which is what you've generally noticed under #159?). 

Mostly note it occurs on ```.plot_centroids()``` method where the legend is quite wide to start with.

Above example is running ```03_custom.py``` example on v3.5.1 of matplotlib with python 3.8.10.

I've seen a few examples with certain wider than height geometries cutting off much more of the legend. Generally if the returned figure is even marginally larger it seems to format better with the wider legends.

robbievanleeuwen:
Hi @Agent6-6-6, sorry for the radio silence on this one. These plotting issues have genuinely stumped me, I'm not sure what I'm doing wrong but feel like it's a `matplotlib` issue. `tight_layout()` is supposed to fix this but they do say:

>tight_layout automatically adjusts subplot params so that the subplot(s) fits in to the figure area. This is an experimental feature and may not work for some cases. It only checks the extents of ticklabels, axis labels, and titles.

So probably an issue with the legends?

For now, I feel a bit defeated with trying to fix this one so would welcome any new & fresh ideas & PRs that look to fix this :pray:

PR #235, opened 2022-07-31
Fix legend formatting

Agent6-6-6:
@robbievanleeuwen, I believe this addresses #188 

I've only tested it briefly on a few examples where I previously had legend formatting issues since it was so intermittent

But I think the issue was calling `tight_layout()` after setting the aspect ratio for the plot did something weird for certain aspect ratio plots. Calling it beforehand seems to work as intended.

### DIFF
--- a/sectionproperties/post/post.py
+++ b/sectionproperties/post/post.py
@@ -7,9 +7,7 @@ from sectionproperties.pre.pre import DEFAULT_MATERIAL
 
 
 @contextlib.contextmanager
-def plotting_context(
-    ax=None, pause=True, title="", filename="", render=True, axis_index=None, **kwargs
-):
+def plotting_context(ax=None, pause=True, title="", filename="", render=True, axis_index=None, **kwargs):
     """Executes code required to set up a matplotlib figure.
 
     :param ax: Axes object on which to plot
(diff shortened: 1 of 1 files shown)


robbievanleeuwen (changes requested):
Hi @Agent6-6-6 thanks for this, it's certainly an improvement on what's currently happening.

In some cases this cuts off the title of the plots, see some of the [examples](https://sectionproperties--235.org.readthedocs.build/en/235/sphinx_gallery_examples/index.html) in the gallery that are built from this pull request. I think overall this is still better than what is currently happening.

Could I get you to run ``black`` on the code so that its correctly formatted and I will merge.

Cheers!

Agent6-6-6:
@robbievanleeuwen, are you just using the default line length in black (88 I think)?

Forgot I have it set to a slightly longer line length which reformatted a few longer lines, but will change and resubmit. 

Hadn't noticed the title getting cutoff, will have a further play around after merging and see if that can be figured out.

Agent6-6-6:
@robbievanleeuwen, 5a336abe18074b8a217c82c08e318322229b27d6 seems to fix the title cutting off on the examples. Just needed the title created prior to `tight_layout()`.

Should be good to merge once the tests pass

robbievanleeuwen:
Great, thanks @Agent6-6-6!

PR #243, opened 2022-08-22
Loosen warping reqs for stress analysis

robbievanleeuwen:
Only require a warping analysis to be performed for a stress analysis if the shear force or twisting moment is non-zero.

### DIFF
--- a/docs/source/rst/analysis.rst
+++ b/docs/source/rst/analysis.rst
@@ -71,8 +71,9 @@ A stress analysis calculates the section stresses arising from a set of forces
 and moments. Executing this method returns a :class:`~sectionproperties.analysis.section.StressResult`
 object which stores the section stresses and provides stress plotting functions.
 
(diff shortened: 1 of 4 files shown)


robbievanleeuwen:
@connorferster wondering if you've had a chance to look at this and whether or not it's fit for purpose for your use case?

connorferster (approved):
Looks good to me! Are you up for adding a test for triggering the `RuntimeError`s?

connorferster:
Woo woo! We are up over 59.5% for code coverage! 60%, here we come! :)

robbievanleeuwen:
😆 setting myself the goal of improving the test coverage over the coming weeks, we'll see how we go!

Thinking about doing a bit of a refactor as well as switching the docs from RTD over to GH pages, similar to *concreteproperties*. Will break a few links in the github repo which I'll try to patch but I think it will be an overall improvement.

PR #245, opened 2022-09-16
add option for side bars to rect concrete section

Agent6-6-6:
@robbievanleeuwen - Added option for side bars to the rectangular section for creating column sections with side bars or deeper beams where side bars are required for crack control.


robbievanleeuwen:
@Agent6-6-6 looks good, thanks!

Just a heads up, if using *concreteproperties* this can be done with [``add_bar_rectangular_array()`` ](https://robbievanleeuwen.github.io/concrete-properties/rst/geometry.html#add-bar-rectangular-array) and setting ``exterior=True``, but only if all the bars are the same size!

robbievanleeuwen (approved):

PR #250, opened 2022-10-17
Geom instantiation fix

connorferster:
I was working away on building a concrete section in `concreteproperties` but kept running into an error saying that I could not "perform difference between my two geometries", whenever I tried to add my bar array to my geometry. The error in `CompoundGeometry.__sub__()` was being caused by an error in `Geometry.__init__` when a geometry is being instantiated with a control point that is an instance of `shapely.geometry.Point`. This occurs when a control point is an assigned control point. However, `Geometry.__init__` did not have a way of checking for this condition. Now it does.

This fixes the error and all existing tests pass. I don't know how to write a test for this fix specifically because it is just addressing expected internal behaviour.

### DIFF
--- a/sectionproperties/pre/geometry.py
+++ b/sectionproperties/pre/geometry.py
@@ -53,7 +53,7 @@ class Geometry:
         self,
         geom: shapely.geometry.Polygon,
         material: pre.Material = pre.DEFAULT_MATERIAL,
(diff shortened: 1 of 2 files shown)


connorferster:
Crap. A whole bunch of code was included from a commit I was working on (code in fd62fad). I will delete this PR and re-submit.

PR #251, opened 2022-10-17
Added check for when control_point is an instance shapely.geometry.Point

connorferster:
Fixes an error that sometimes comes up during difference operations. The error states that `Cannot perform difference between these two geometries...` and it is thrown because another error is caused during instantiation of the `Geometry` instance. The error occurs because the value of `control_point` being passed is an instance of `shapely.geometry.Point` instead of the assumed `list[float, float]`. This condition occurs when a control_point has been assigned and is being carried through an operation. 

This PR acknowledges that a control_point of type `shapely.geometry.Point` is a valid value and updates the function signature and check to match.

### DIFF
--- a/sectionproperties/pre/geometry.py
+++ b/sectionproperties/pre/geometry.py
@@ -53,7 +53,7 @@ class Geometry:
         self,
         geom: shapely.geometry.Polygon,
         material: pre.Material = pre.DEFAULT_MATERIAL,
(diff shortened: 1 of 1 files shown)


connorferster:
Codecov: I typically write tests but did not know how to write a test for this particular fix because it is updating internal behaviour to be self-consistent.

robbievanleeuwen (approved):

PR #252, opened 2022-10-18
Additional sections for concrete_sections

connorferster:
This PR adds two new geometry generating functions to the `concrete_sections` module:
1. `concrete_column_section`
2. `add_bar` - Previously only in `concreteproperties`

As I have been working with creating concrete sections, I found a need for including the incredibly useful `add_bar` function into `sectionproperties`. I find that it is a general-purpose function that can be utilized whenever you are reading bar locations from an external data source (e.g. a text file). It allows you to treat your concrete section as an accumulator and for bars to easily be added in a loop.

Additionally, I think it would be useful to include a function for creating a rectangular concrete section with a regular bar array where all bars are the same size and diameter. I called this function `concrete_column_section` to differentiate it from `concrete_rectangular_section` since the orientation of the bars in such arrays would be most like those used in typical concrete columns.

I am open to changing the name of `concrete_column_section` but am submitting it for review and consideration.

### DIFF
--- a/sectionproperties/pre/library/concrete_sections.py
+++ b/sectionproperties/pre/library/concrete_sections.py
@@ -1,4 +1,4 @@
-from typing import Union
+from typing import Union, Optional
 import numpy as np
 from shapely.geometry import Polygon
 import sectionproperties.pre.pre as pre
@@ -133,11 +133,11 @@ def concrete_rectangular_section(
 def concrete_column_section(
     b: float,
     d: float,
-    dia_bar: float,
-    bar_area: float,
     cover: float,
     n_bars_b: int,
     n_bars_d: int,
+    dia_bar: float,
+    bar_area: Optional[float] = None,
     conc_mat: pre.Material = pre.DEFAULT_MATERIAL,
     steel_mat: pre.Material = pre.DEFAULT_MATERIAL,
     filled: bool = False,
@@ -150,11 +150,13 @@ def concrete_column_section(
 
     :param float b: Concrete section width, parallel to the x-axis
     :param float d: Concrete section depth, parallel to the y-axis
-    :param float dia_bar: Diameter of reinforcing bars. Used only for calculating bar location.
-    :param float bar_area: Area of reinforcing bars. Used for section capacity calculations.
     :param float cover: Clear cover, calculated as distance from edge of reinforcing bar to edge of section.
     :param int n_bars_b: Number of bars placed across the width of the section, minimum 2.
     :param int n_bars_d: Number of bars placed across the depth of the section, minimum 2.
+    :param float dia_bar: Diameter of reinforcing bars. Used for calculating bar placement and,
+        optionally, for calculating the bar area for section capacity calculations.
+    :param Optional[float] bar_area: Area of reinforcing bars. Used for section capacity calculations.
+        If not provided, then dia_bar will be used to calculate the bar area.
     :param Optional[sectionproperties.pre.pre.Material] conc_mat: Material to
         associate with the concrete
     :param Optional[sectionproperties.pre.pre.Material] steel_mat: Material to
@@ -237,6 +239,8 @@ def concrete_column_section(
         xy = np.meshgrid(b_edge_bars_x, d_edge_bars_y)
         all_bar_coords = np.append(xy[0].reshape(-1, 1), xy[1].reshape(-1, 1), axis=1)
 
+    if bar_area is None:
+        bar_area = np.pi * dia_bar**2 / 4
     for bar_coord in all_bar_coords:
         concrete_geometry = add_bar(
             concrete_geometry,
--- a/sectionproperties/pre/library/concrete_sections.py
+++ b/sectionproperties/pre/library/concrete_sections.py
@@ -167,7 +167,7 @@ def concrete_column_section(
         number of points used in the circle. Useful for making the reinforcing bars look
         more circular when plotting the concrete section.
 
-    :raises ValueErorr: If the number of bars in either 'n_bars_b' or 'n_bars_d' is not greater
+    :raises ValueError: If the number of bars in either 'n_bars_b' or 'n_bars_d' is not greater
         than or equal to 2.
 
     The following example creates a 600D x 300W concrete column with 25 mm diameter
--- a/sectionproperties/pre/library/concrete_sections.py
+++ b/sectionproperties/pre/library/concrete_sections.py
@@ -145,8 +145,8 @@ def concrete_column_section(
 ) -> geometry.CompoundGeometry:
     """Constructs a concrete rectangular section of width *b* and depth *d*, with
     steel bar reinforcing organized as an *n_bars_b* by *n_bars_d* array, discretised 
-    with *n_circle* points with equal side and top/bottom
-    *cover* to the steel.
+    with *n_circle* points with equal sides and top/bottom *cover* to the steel which
+    is taken as the clear cover (edge of bar to edge of concrete).
 
     :param float b: Concrete section width, parallel to the x-axis
     :param float d: Concrete section depth, parallel to the y-axis
@@ -170,10 +170,11 @@ def concrete_column_section(
     :raises ValueErorr: If the number of bars in either 'n_bars_b' or 'n_bars_d' is not greater 
         than or equal to 2.
 
-    The following example creates a 600D x 300W concrete beam with 3N20 bottom steel
-    reinforcing bars and 30 mm cover::
+    The following example creates a 600D x 300W concrete column with 25 mm diameter
+    reinforcing bars each with 500 mm**2 area and 35 mm cover in a 3x6 array without
+    the interior bars being filled::
 
-        from sectionproperties.pre.library.concrete_sections import concrete_rectangular_section
+        from sectionproperties.pre.library.concrete_sections import concrete_column_section
         from sectionproperties.pre.pre import Material
 
         concrete = Material(
@@ -186,8 +187,8 @@ def concrete_column_section(
         )
 
         geometry = concrete_column_section(
-            b=300, d=600, bar_diam=dia_top=20, n_top=0, dia_bot=20, n_bot=3, n_circle=24, cover=30,
-            conc_mat=concrete, steel_mat=steel
+            b=300, d=600, dia_bar=25, bar_area=500, cover=35, n_bars_b=3, n_bars_d=6, 
+            conc_mat=concrete, steel_mat=steel, filled=False, n_circle=4
         )
         geometry.create_mesh(mesh_sizes=[500])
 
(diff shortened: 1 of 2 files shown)


connorferster:
Woot woot! We broke the 60% code coverage mark!

Agent6-6-6:
@robbievanleeuwen, @connorferster one thing I've been meaning to raise on these methods is why the bar area is not derived by default from the bar diameter. I see no practical use for not doing this if the intent is you're using these section generation methods with `concreteproperties` to determine a capacity.

I just find by default I constantly have to define the bar diameter, and determine the bar area based on this as an extra potentially unnecessary step:-
```
bar_dia = 28

.... dia_top=bar_dia, area_top=bar_dia**2 * np.pi / 4, ....
```
This additional step to calculate `area_top` in the example as an input seems a little redundant to me when this could just be done within the method itself by default. 

I'm probably missing something here regarding why there is potential for defining bar size and area independently?

sectionproperties/pre/library/concrete_sections.py
```diff
@@ -147,27 +148,27 @@ def concrete_column_section(
...
+        associate with the reinforcing steel
+    :param bool filled: When True, will populate the concrete section with an equally
+        spaced 2D array of reinforcing bars numbering 'n_bars_b' by 'n_bars_d'. 
+        When False, only the bars around the perimeter of the array will be present.
+    :param int n_circle: The number of points used to discretize the circle of the reinforcing
+        bars. The bars themselves will have an exact area of 'bar_area' regardless of the
+        number of points used in the circle. Useful for making the reinforcing bars look
+        more circular when plotting the concrete section.
+
+    :raises ValueErorr: If the number of bars in either 'n_bars_b' or 'n_bars_d' is not greater
```
Agent6-6-6: @connorferster ValueError spelt incorrectly here

sectionproperties/pre/library/concrete_sections.py
```diff
@@ -170,10 +170,11 @@ def concrete_column_section(
     :raises ValueErorr: If the number of bars in either 'n_bars_b' or 'n_bars_d' is not greater 
         than or equal to 2.
 
-    The following example creates a 600D x 300W concrete beam with 3N20 bottom steel
-    reinforcing bars and 30 mm cover::
+    The following example creates a 600D x 300W concrete column with 25 mm diameter
+    reinforcing bars each with 500 mm**2 area and 35 mm cover in a 3x6 array without
```
Agent6-6-6, line 174: for this example should the bar area be corrected to 491mm^2 for a 25 mm diameter metric bar?

robbievanleeuwen:
Hi @Agent6-6-6 I guess it's more been kept as a legacy feature as circles were originally defined only by diameter in ``sectionproperties``. 

Inferring the diameter from the area would work fine, the only issue would be a slight misalignment with the positioning of the bars. For example in calculating the distance from the top of a beam (say at ``y=0``) to the centre of an N16 bar with 30 mm cover:

1. **Current implementation:** position is determined based on ``diameter=16`` and would give ``y=30+16*0.5=38`` to the centre of the circle.
2. **Proposed implementation:** design area of bar is provided ``area=200``, inferred diameter would be different to ``16`` (slightly larger but dependent on number of discretisation points), position would be ``y>38``.

Agent6-6-6:
Don't think I've been clear. I'm after option 1, but instead of having the area as a separate input. Why isn't it just calculated from the diameter supplied by user and then worked back to an appropriate polygon with the same area for creating the section?

robbievanleeuwen:
The reason I have kept it is that, in Australia at least (I believe), the "calculated area" is not equal to the "nominal area", that is the area used for design. See below:

Screen Shot 2022-10-18 at 9 15 42 pm

This results in a (small) discepancy between the diameter and the area. 

But it may be worth just deciding that the difference is insignificant and that we should scrap ``diameter`` all together for simplicity and avoiding repetition.

Thoughts @Agent6-6-6 and @connorferster?

connorferster:
It is the same in CSA where both the nominal bar diameters and areas are different than the actual. While, like @Agent6-6-6, I am generally in favour of reducing the number of inputs to enhance ease of use, I find that for these applications, having two separate inputs allows for greater control and less surprises in the results. I think specific code implementations can reduce these to a single input, like bar_size, where knowledge of the nominal diameter and area can be provided to both inputs and some value like "25M" can be provided instead.

connorferster:
In the test that I wrote, I explicitly take advantage of this decoupling of diameter and are for testing the resulting position of the bars.

Agent6-6-6:
And explain in docs that area is only required if you want a nominal bar area, otherwise actual bar area is calculated.

sectionproperties/pre/library/concrete_sections.py
connorferster: Thanks! Will fix

sectionproperties/pre/library/concrete_sections.py
connorferster: I am intending to use both a nominal diameter and nominal area to demonstrate that the area and diameter inputs are decoupled.

connorferster:
My preference is to leave it as is for the following reasons:

1. Having decoupled inputs allows for any use case: diam and area both using nominal values, two different values e.g. area using actual and diameter using nominal, or both values being actual.
2. Deriving values from the discretized circle creates unexpected values that vary with the number of points
3. Introducing an independent floating point approximation of an analytical calculation for one or both values introduces (i.e. pi*d**2/4) a third source of data and potentially unexpected result: is my bar diameter based off of my bar discretization, some nominal value, or some independent calculation?
4. Keeping it as is consistent with the other concrete functions which offer the same flexibility.

I think adding these input options into each base concrete geometry generation function creates more code and more documentation than I think is necessary. The easiest input case would be managed through a code implementation with a bar_size and possibly a boolean to indicate actual vs nominal.

robbievanleeuwen:
@connorferster just a heads up, the existing concrete sections follow @Agent6-6-6's comment, where the parameter ``area_bar: float = None``. If not provided, the area is constructed from a circle given the diameter, otherwise the area and the number of points defines the circle and the diameter determines the position.

I think this allows for the flexibility of allowing the using to enter an area that doesn't *mathematically* correspond to the diameter.

I think that changing the implementation so that ``area_bar: float = None`` results in the area being calculated based off the discretisation points is a better approach. Currently the implementation can lead to users unknowingly having a different bar area than what they might expect due to discretisation errors.

This let's users not have to calculate the area themselves if they don't want to (extra work, extra chance for errors).

connorferster:
> @connorferster just a heads up, the existing concrete sections follow @Agent6-6-6's comment, where the parameter `area_bar: float = None`. If not provided, the area is constructed from a circle given the diameter, otherwise the area and the number of points defines the circle and the diameter determines the position.

Lol, right! I see that now. My bad. 

That being the case, I would _further_ make the recommendation of removing the optional arguments in all other concrete functions and requiring both diameter and area for all. This would reduce the testing surface, complexity, and potentially fix any buggy behaviour in the future.

@robbievanleeuwen If the discretized bar is used to calculate area, then I think it will lead to unexpected results with bar reinforcing ratios being very different than the expected. I have created this gist as a demonstration: https://gist.github.com/connorferster/4f5749544ed95bff3c374de00249b02c

The reinforcing ratio is more than a whole percentage point different when comparing the two ways of entering bar size.

While this _could_ be documented behaviour, my thoughts are, "why bother with these potential errors?". If a user is entering section information into concreteproperties manually then one extra input out of several inputs is not a big deal. If the user is parsing data and then creating sections automatically then the extra input is trivial.

I think that, for the ability to always produce expected results, to reduce the testing surface, to reduce the potential for bugs in future functions, and to reduce the complexity of functions and their documentation, it would be _advantageous_ to remove the optional arguments and require bar diameter and bar area be entered for all concrete section geometry generation functions.

Agent6-6-6:
This is kind of my point, instead of doing whats currently being done do this if no nominal area is provided otherwise you have the exact issue highlighted in your gist.

```
# work out true area
area_top = dia_top**2np.pi/4
# then do the by area thing to maintain the correct area based on diameter instead of jumping the diameter and n_points which will always underestimate the area!
bar = primitive_sections.circular_section_by_area(
                area=area_top, n=n_circle, material=steel_mat
```

Agent6-6-6:
> can specify diameter only, then actual area is calculated.
Actual area is not calculated at the moment.

> **the existing concrete sections follow @Agent6-6-6's comment,**

At the moment this is not the case, it's creating a section for the bar based on the exterior diameter based on n_points which will always underestimate the area, sometimes by a significant margin unless n_points is set to a significantly high value to converge on the actual area (which the default of 4 does not achieve..

robbievanleeuwen:
Sorry for the confusion on this thread! For clarity, are we all in agreement with the following?

```python
def generic_concrete_section(
    geom: Any,  # generic concrete geometry
    dia: float,
    cover: float,
    n_bars: float,
    conc_mat: pre.Material = pre.DEFAULT_MATERIAL,
    steel_mat: pre.Material = pre.DEFAULT_MATERIAL,
    area_bar: Optional[float] = None,
    n_circle: int = 4,
) -> geometry.CompoundGeometry:

    # do stuff with geometry...

    # add reinforcement
    if not area_bar:
        area_bar = np.pi * dia * dia / 4

    bar = primitive_sections.circular_section_by_area(
        area=area_bar, n=n_circle, material=steel_mat
    )  # NEVER use primitive_sections.circular_section() for bars!

    # shift bar as required using geom, dia and cover

    geom = (geom - bar) + bar

    return geom
```

This way there area never any *incorrect* areas being applied to the bars. If not please feel free to modify the above!

Agent6-6-6:
@robbievanleeuwen, sounds good to me as a general approach to apply to all similar methods.

connorferster:
@robbievanleeuwen @Agent6-6-6 

Ok, agreed. I will implement it in this pull request.

connorferster:
@robbievanleeuwen Let me know if this is good to go!

Agent6-6-6:
> @robbievanleeuwen Let me know if this is good to go!

@connorferster I think all the other original methods should have the Optional type like the new method you've added??
`area: Optional[float] = None` vs `area: float = None`

connorferster:
Got it. Will update

sectionproperties/pre/library/concrete_sections.py
```diff
@@ -33,12 +36,17 @@ def concrete_rectangular_section(
...
         circle discretisation)
     :param float area_bot: If provided, constructs bottom reinforcing bars based on
         their area rather than diameter (prevents the underestimation of steel area due
         to circle discretisation)
+    :param float area_side: If provided, constructs side reinforcing bars based on
+        their area rather than diameter (prevents the underestimation of steel area due
+        to circle discretisation)
     :param Optional[sectionproperties.pre.pre.Material] conc_mat: Material to
         associate with the concrete
     :param Optional[sectionproperties.pre.pre.Material] steel_mat: Material to
```
Agent6-6-6, line 52: the ValueError below this has a similar spelling mistake to that noted on added column section

Agent6-6-6: same also in `concrete_tee_section` method.

connorferster:
@robbievanleeuwen Are we good to go on this one? I will fix the spelling mistakes later but your recent changes require some fussing on my end on my local. Will be easier to fix after merging.

connorferster:
@robbievanleeuwen My bad. No changes from your end. My local branch is saying the branches have diverged (?!) Do not know why. However, this PR has no conflicts. Am hoping to merge and fix the propagated spelling mistake afterwards.

robbievanleeuwen (approved):

PR #262, opened 2023-01-02
Unit test stress at point

normanrichardson:
These tests are intended to extend the testing of the get_stress_at_points functionality. They are based on the following simple cases: 

- rectangular section with a mxx applied
- rotated rectangular section such that mxx is applied about the diagonal


robbievanleeuwen (approved):
Thanks for this @normanrichardson :)

Issue #266, opened 2023-02-10
Function Vx shear stress plot with incorrect title


Chivos:
Hi,

I noticed something weird at this line :
https://github.com/robbievanleeuwen/section-properties/blob/12763207b0981c9932cbe1b56914b84057ce3491/sectionproperties/analysis/section.py#L3632
In this function for plotting Vx shear stress, there is a title refering to the tension caused by bending.

I found this while trying to understand why I'm unable to give my own title on the plot for Vx shear in a 3x3 subplot...
Didn't find for now if it's my code or sectionproperties...

Chivos:
Seems to be in conflict with line 3672

robbievanleeuwen:
Hi @Chivos thanks for this, will merge the PR.

PR #267, opened 2023-02-11
Update section.py - Issue  #266

Chivos:
This should solve issue #266


robbievanleeuwen:
Thanks again @Chivos.

Note that I will soon be releasing v3.0.0 which will change the way plots are generated. As a result, I will not be releasing a new version that includes these changes, but these changes can still be used by cloning the repo and installing manually.

PR #274, opened 2023-05-05
Pass through DXF import options for better geometry creation

ccaprani:
For curves and other complex geometries, the default options in `cad2shapely` may not suffice. This PR exposes the optional inputs to the `sectionproperties` user in `from_dxf`.

A few minor tweaks to the documentation for correctness, and black autoformatting removed a few parentheses.

Issue #272, opened 2023-04-28
Unexpected behavior calculating 2nd Moments of Area (.get_ic) with Material


aedificatio:
Hi,  First a big complement for the great work that has been done with sectionproperties.

**Describe the bug**
Unexpected results for `.get_ic()` when using materials.
Both the documentation and the Docstring advertise the result being the second moments of area around the centroidal axis. There is NO mention of weighted results by materials.
The function returns I (ixx, iyy, ixy) when no material is applied and E times I (EI) when materials are applied. In my opinion this is inconsistent and unexpected behavior.

**Expected results**
`.get_ic()` always returns the second moments of area based on the geometry

**Additional context**
For a compound section the most accurate parameter would be EI. The concept of weighted results only relates to a compound section which includes different materials. The user could simply compute Eeff or Ic;eff.
I think the results of `.get_ic()` should only return the second moments of area based on the geometry. If the result is ambiguous because of different materials it could raise a warning, but still compute Ic based on the geometry.
In https://sectionproperties.readthedocs.io/en/latest/rst/post.html#how-material-properties-affect-results it is explained in more detail. The issue (or related) is also mentioned in #6 #16 #76 and #207. As a workaround an E-mod of 1 could be used or no material at all.

**Proposed solution**
*	`.get_ic()` only returns the geometric values of a (compound) section regardless of material
*	if different materials are being applied maybe raise a warning but still compute Ic
*	implement something like `.get_EI()` or `.get_EIeff()` to get the weighted result of a (compound) section

if the above is not implemented then please adjust the documentation to reflect the real behavior.

Kind regards.

ccaprani:
Hi @aedificatio ; not @robbievanleeuwen, but a contributor & interested engineer. I suspect you haven't gotten much response because the tone of your message comes across a little strong; I'm sure you didn't mean it that way.

It's certainly not dangerous to return material-weighted section properties. Indeed, this is the expected behaviour for any structural engineering application I can think of. 

Is there a structural engineering use-case for returning the unweighted properties for a composite section?

Nevertheless, if a user requires the geometric properties of a composite section without reference to any material weighting, then the materials can be set equal; or a nominal material of `E = 1` can be used, as you note. I wouldn't describe this as a workaround; it's just using the software to obtain the results of interest.

I think the key point is that the software is not inherently dangerous, and it's reasonable to expect users to know how to interpret the results. Adding warnings for regular use-cases then seems unnecessary.

[...]

Issue #268, opened 2023-02-12
section.calculate_warping_properties(solver_type='cgs') fails due to NameError


simo-11:
**Describe the bug**
Calculation of warping properties fails when using solver_type='cgs'

**To Reproduce**
section.calculate_warping_properties(solver_type='cgs')

sectionproperties\analysis\section.py", line 402, in ilu_decomp
    (self.num_nodes, self.num_nodes), linalg.spilu(k).solve

NameError: name 'k' is not defined

**Additional context**
It seems that issue due partially made refactoring as assemble_torsion was modified to return only lagrangian version of k (k_lg) and no k at all.

Fixing
 * remove usage of k  and k_precond in calculate_warping_properties will be provided

robbievanleeuwen:
Thanks for picking this up @simo-11. I have resolved this for the upcoming release, see [here](https://github.com/robbievanleeuwen/section-properties/blob/hypermodern/src/sectionproperties/analysis/section.py#L393).

Fixed by #277.

Issue #259, opened 2022-12-16
Better mesh size control in triangle


robbievanleeuwen:
### Discussed in https://github.com/robbievanleeuwen/section-properties/discussions/257

<div type='discussions-op-text'>

<sup>Originally posted by **audunarn** December  2, 2022</sup>
Hi, I find that it is easier to control the mesh size in triangle by changing the max/min angle (in pre.py):

# generate mesh
if coarse:
    mesh = triangle.triangulate(tri, "pAo2")
else:
    mesh = triangle.triangulate(tri, "pq**30**Aao2")

I could add this functionality if you wish.</div>

SoundsSerious:
I think this would really help with some of the segmentation fault issues #284 that I find will occur sometimes.

Would there be any advantages to trying a different meshing library?

robbievanleeuwen:
Hi @SoundsSerious, there is a new library [cytriangle](https://github.com/m-clare/cytriangle) in development that looks promising - watch this space.

SoundsSerious:
Yes that looks great, hopefully that would plug right into `sectionproperties`

PR #277, opened 2023-06-10
v3.0

robbievanleeuwen:
# ``sectionproperties`` v3.0

## :boom: Breaking Changes

- Update section property ``get()`` methods behaviour as per #272. The behaviour is unchanged for analyses without material properties, however composite analyses now require material specific get methods to be used.
  - e.g. for a composite analysis, ``get_ic()`` will raise an error telling the user to instead use ``get_eic()``.
- Stress plotting methods have been simplified to improve maintainability and code repetition.
  - Stress plots are now generated by calling the ``StressPost.plot_stress()`` method, with a string determining the type of stress to plot. The previous implementation had a separate method for each different type of stress plot which was very repetitive.
  - There is a similar implementation for generate vector stress plots.
- ``get_stress_at_point()`` method removed, user can use more versatile ``get_stress_at_points()`` method instead.
- Pythonic renaming of methods and variables, e.g. ``get_As()`` becomes ``get_as()`` for retrieving shear areas, moments and shears change from ``Mxx`` and ``Vx`` to ``mxx`` and ``vx``.
- Concrete section library:
  - Added option for variable covers to top, bottom and side reinforcement.
  - Renamed argument names to be more pythonic.
  - Reinforcement now requires an area and diameter (area used to determine discretised circle geometry, diameter used to calculate cover and location).

## :rocket: Features

- Packaging and dependency management with poetry.
- Improve length/readability of imports by placing key classes in ``__init__.py`` files.
  - e.g. ``from sectionproperties.pre.library.primitive_sections import rectangular_section`` becomes ``from sectionproperties.pre.library import rectangular_section`` and ``from sectionproperties.analysis.section import Section`` becomes ``from sectionproperties.analysis import Section`` (note not a breaking feature as previous imports still work).
- #259
- #274 

## :bug: Fixes

- #255
- #268
- #272 

## :rotating_light: Testing

- Use nox & nox-poetry for testing.
- Move ``test_plastic_centroid()`` to ``test_plastic.py``.
- Further TBC.

## :construction_worker: Continuous Integration

- Added dependabot.
- Added labeller and release drafter automation.
- Added pre-commit checks.
- Updated all workflows.

## :books: Documentation

- Changed documentation theme to [furo](https://github.com/pradyunsg/furo).
- General overhaul of the documentation.
- Auto-generate API documentation using sphinx autodoc.
- Auto-generate images in documentation using matplotlib plot directive.
- Change examples from sphinx gallery to notebook style with nbsphinx.
- Add intersphinx mappings.

## :lipstick: Style

- Fully implement typing.
- Added flake8 linter to pre-commit.
- Added darglint docstring linter to pre-commit.
- Added isort to pre-commit.
- Added pyupgrade to pre-commit.
- General refactor to improve readability and maintainability, e.g. reduce amount of code in ``section.py``: ``PlasticSection`` -> ``plastic_section.py``, ``StressPost`` and ``StressResult`` -> ``stress_post.py``, ``SectionProperties`` -> ``post.py``.
- Generally changed type of a point from ``list[float]`` to a ``tuple[float, float]``.

## Other

- Restructured package by adding ``src`` folder.


robbievanleeuwen:
@connorferster thanks for the offer for help - if you have time one thing that would be great would be having a look at #255.

connorferster:
V3.0 looks awesome. Will do #255. Did the upstream issue with import_dxf only returning one single geometry from a DXF file that had many geometries get resolved? If not, can work on that too.

Is the triangle dependency still locking us into Python 3.9?

ccaprani:
Looking good @robbievanleeuwen ! I'd like it if `rhino3dm` could be removed as a dependency & be optional. It's a relatively edge use-case I think, and a risk to long term sustainability...the benefits of 3.11 should be realised by what's mostly a native python application as soon as possible.

robbievanleeuwen:
This is a good point @ccaprani! I'll have a play with around with the dependencies - I'll also have to figure out how to only run the rhino tests for certain versions of python, but I'm pretty sure this is possible with nox!

Spectre5:
It is definitely possible with pytest itself and they examples of similar situations in the docs

robbievanleeuwen:
Thanks @Spectre5, definitely a simpler solution - seems like the [skipif](https://docs.pytest.org/en/7.1.x/reference/reference.html?highlight=skipif#pytest.mark.skipif) marker can do this!

ccaprani:
@robbievanleeuwen if I've understood correctly, can I suggest that cad2shapely and rhino3dm dependencies are seperately optional? I can see the temptation to join them, but I see them as separate use-cases. Preferable not to need rhino3dm just to do a DXF import imo. Even if currently both are limited to 3.10, this could change for one or other.

robbievanleeuwen:
Hi @ccaprani , thanks for the feedback. Yes I agree, this is a good idea to separate the two!

I will create the following extras:

1. ``pip install sectionproperties[dxf]``
2. ``pip install sectionproperties[rhino]``

PR #285, opened 2023-07-19
Allow unlipped cee and zed steel_sections

smith120bh:
Unlipped Cee sections are very common. For example, all track ("T") sections in the Steel Framing Industry Association standard sections, [as defined here](https://sfia.memberclicks.net/assets/TechFiles/sfia%20tech%20spec%202015%20updated%207.12.17.pdf), are unlipped Cee sections. However, `section-properties` currently explicitly throws an error if lip length `l` is set to anything less than `r_out` (the outer radius). 

This pull request adds the ability to set `l` to zero - or anything less than `r_out`. 

Note that I did have a question as to what to do for `0 < l < r_out`. One option would be to throw an error in this case, but I opted instead for decreasing the outer radius commensurately, on the philosophy that fewer error conditions is probably better, but am happy to change this as well. 

While unlipped Zed sections are less common than unlipped Cees, they do exist - and a nearly-identical update could be applied to `zed_section()` to allow unlipped zeds as well here.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,14 @@
 Changelog:
 ==========
 
+v2.1.5:
+-------
+
+- Fix shapely 2.0 imports and STRtree implementation, with thanks to `@normanrichardson <https://github.com/normanrichardson>`_
+- Add support for python 3.10, drop support for python 3.7
+
+**Full changelog:** `2.1.4...2.1.5 <https://github.com/robbievanleeuwen/section-properties/compare/2.1.4...2.1.5>`_
+
 v2.1.4:
 -------
 
--- a/sectionproperties/__init__.py
+++ b/sectionproperties/__init__.py
@@ -6,4 +6,4 @@ sectionproperties can be used to determine section properties to be used
 in structural design and visualise cross-sectional stresses resulting
 from combinations of applied forces and bending moments.
 """
-__version__ = "2.1.4"
+__version__ = "2.1.5"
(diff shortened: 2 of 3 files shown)


robbievanleeuwen:
Hi @smith120bh, thanks for this PR and adding this functionality. Overall looks pretty good, I will review shortly and look to merge into the `hypermodern` (dev) branch. I'll also have a closer look at the edge case you mentioned and let you know what I think!

robbievanleeuwen:
Hi @smith120bh becuase this PR was forked from the `master` branch, merging into `hypermodern` is proving quite tricky, sorry about that! I think because the changes are quite localised, the best way forward would be to create a new branch from `hypermodern` and copy-paste these changes then create a new PR from there.

If you don't have time at the moment I am happy to make these changes, however you wouldn't be attributed the changes in `git` 😆 I could always add a comment in the code for attribution. Let me know how you think is best to proceed, cheers!

connorferster:
I like the implementation (`if l < t`). This seems like a really useful feature.

smith120bh:
@robbievanleeuwen - Sure - I'll recreate the PR off of the `hypermodern` branch momentarily!

smith120bh:
Mon dieu - it took me forever to realise that there's a _GITHUB_ setting when you first fork a repository for 'copy only the `master` branch', which then made `git fetch upstream; git checkout -b hypermodern upstream/hypermodern` error out. Now working.

At any rate, @robbievanleeuwen , now recreated as PR #300 ! I think I converted everything to your new types and sp_utils function, but you should probably double-check :-)

robbievanleeuwen:
Thanks @smith120bh, again sorry for the branch confusion! I'll review shortly :)

PR #297, opened 2023-09-30
Proposed fix for non-recognized holes

connorferster:
When the geometries resulting from an operation result in a MultiPolygon that only touch at points, a hole is not recognized in shapely but it is "recognized" by humans. To fix this problem, the hole needs to be recognized by shapely as part of a Polygon. One solution to fix this would be to buffer the geometry by a very small amount to force this switch from MultiPolygon to Polygon.

### DIFF
--- a/src/sectionproperties/pre/geometry.py
+++ b/src/sectionproperties/pre/geometry.py
@@ -25,6 +25,8 @@ import sectionproperties.post.post as post
 import sectionproperties.pre.bisect_section as bisect
 import sectionproperties.pre.pre as pre
 
+SCALE_CONSTANT = 1e-9
+
 
 class Geometry:
     """Class for defining the geometry of a contiguous section of a single material.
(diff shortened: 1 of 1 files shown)


robbievanleeuwen:
Looks good @connorferster, thanks! Just a small fix required by `flake8` to be more pythonic:

```
src/sectionproperties/pre/geometry.py:2387:10: N806 variable 'SCALE_CONSTANT' in function should be lowercase
```

Will merge after this change :)

src/sectionproperties/pre/geometry.py
```diff
@@ -2384,7 +2384,10 @@ def compile_geometry(self) -> None:
         # Determine if new holes have been created or if existing
         # holes have been destroyed (or "filled in").
         resultant_holes = []
+        SCALE_CONSTANT = 1e-9
```
robbievanleeuwen, line 2387: `scale_constant = 1e-9`

robbievanleeuwen (changes requested):
Looks good @connorferster, thanks! Just a small fix required by flake8 to be more pythonic:
```
src/sectionproperties/pre/geometry.py:2387:10: N806 variable 'SCALE_CONSTANT' in function should be lowercase
```
Will merge after this change :)

connorferster:
Hmm...maybe I should put the scale constant outside of the function then? As per PEP8, constants should be all caps. I will try moving it to the module level (which I prefer not to do because then it will be visible during the module imports) to see if that makes flake8 happy.

robbievanleeuwen:
Thanks @connorferster that should probably work - note `N806` is "variable in function should be lowercase"

PEP8 says "Constants are usually defined on a module level", I guess that's what is making the linter upset.

robbievanleeuwen (approved):

PR #298, opened 2023-09-30
More improvement

TLCFEM:
Use 4-point quadrature. Other small improvements.

### DIFF
--- a/src/sectionproperties/analysis/fea.py
+++ b/src/sectionproperties/analysis/fea.py
@@ -97,10 +97,9 @@ class Tri6:
 
         # loop through each Gauss point
         for gp in gps:
-            # determine shape function, shape function derivative and jacobian
-            n, _, j = shape_function(coords=self.coords, gauss_point=gp)
-
-            nx, ny = self.coords @ n
+            # determine shape function, shape function derivative,
+            # jacobian and global coordinates
+            n, _, j, nx, ny = shape_function(coords=self.coords, gauss_point=gp)
 
             weight = gp[0] * j
 
@@ -137,11 +136,9 @@ class Tri6:
         gps = gauss_points(n=4)
 
         for gp in gps:
-            # determine shape function, shape function derivative and jacobian
-            n, b, j = shape_function(coords=self.coords, gauss_point=gp)
-
-            # determine x and y position at Gauss point
-            nx, ny = self.coords @ n
+            # determine shape function, shape function derivative,
+            # jacobian and global coordinates
+            n, b, j, nx, ny = shape_function(coords=self.coords, gauss_point=gp)
 
             weight = gp[0] * j * self.material.elastic_modulus
 
@@ -177,11 +174,9 @@ class Tri6:
         gps = gauss_points(n=4)
 
         for gp in gps:
-            # determine shape function, shape function derivative and jacobian
-            n, b, j = shape_function(coords=self.coords, gauss_point=gp)
-
-            # determine x and y position at Gauss point
-            nx, ny = self.coords @ n
+            # determine shape function, shape function derivative,
+            # jacobian and global coordinates
+            n, b, j, nx, ny = shape_function(coords=self.coords, gauss_point=gp)
 
             weight = gp[0] * j * self.material.elastic_modulus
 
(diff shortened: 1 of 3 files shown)


src/sectionproperties/analysis/fea.py
```diff
@@ -92,8 +92,8 @@ def geometric_properties(
         iyy: float = 0.0
         ixy: float = 0.0
 
-        # Gauss points for 6 point Gaussian integration
-        gps = gauss_points(n=6)
+        # Gauss points for 4 point Gaussian integration
+        gps = gauss_points(n=4)
```
TLCFEM, line 96: For T6, 3-point quadrature (2nd order) is sufficient. 6-point quadrature (5th order) is an overkill. Here use 4-point (3rd order) should be okay to cover most cases, even with nonlinear material. This reduces 30% time spent on elemental quantities.

src/sectionproperties/analysis/solver.py
```diff
@@ -118,9 +118,10 @@ def solve_direct_lagrange(
     u = spsolve(A=k_lg, b=np.append(f, 0))
 
     # compute error
-    err = u[-1] / max(np.absolute(u))
+    multiplier = abs(u[-1])
```
TLCFEM, line 121: Need to test both absolute and relative error. If the RHS is trivial, the solution will be close to epsilon, making the relative error randomly large. This causes random failure of tests on different machines.

src/sectionproperties/analysis/fea.py
```diff
@@ -900,31 +835,30 @@ def __shape_function_cached(
         dtype=float,
     )
 
+    coords_array = np.array(coords).reshape((2, 6))
```
TLCFEM, line 838: See https://perfpy.com/472

tests/test_rectangle.py
```diff
@@ -118,7 +118,9 @@ def test_rectangular_section_plastic():
 def test_rectangular_section_warping():
     """Test rectangular section warping properties."""
     check.almost_equal(rectangle_section.section_props.j, 2861002, rel=2 * warp_tol)
-    check.almost_equal(rectangle_section.section_props.gamma, 3.177234e08, rel=tol)
+    check.almost_equal(
+        rectangle_section.section_props.gamma, 3.177234e08, rel=5 * warp_tol
```
robbievanleeuwen, line 122: This difference is puzzling, all other warping results appear to be exactly the same with your changes.

I've narrowed this down to changing `n=6` to `n=4` Gauss points in `shear_warping_integrals()`. If I'm being honest, it's a very long time since I've thought about quadrature requirements given the functions we're integrating (5+ years?) I'm wondering what your thoughts are on this one? 

Do you think `n=4` is sufficient given that the results for `gamma` have changed? Given my *current* lack of insight on this I'm hesitant on making any changes that break the tests results and am leaning towards keeping `n=6` here.

robbievanleeuwen (commented):
Thanks for these extra changes @TLCFEM, I have left a quick comment regarding the change in the warping constant result (gamma).

src/sectionproperties/analysis/fea.py
```diff
@@ -790,7 +722,7 @@ def local_coord(
 
 
 @lru_cache(maxsize=None)
-def gauss_points(n: int) -> np.ndarray:
+def gauss_points(*, n: int) -> np.ndarray:
```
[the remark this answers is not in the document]
robbievanleeuwen: Disregard, just figured out what this is!

tests/test_rectangle.py
TLCFEM: Theoretically, the shape function used in T6 is complete up to second order, which should be accurately integrated using a 3-point quadrature. In practical FEA, considering material nonlinearity, often a 4-point quadrature is chosen for T6, which is enough, there is no need to use higher order one because when material nonlinearity is involved, both quadrature won't be accurate/exact. Mesh refinement is require anyway for better results.

In this particular case, integrating N*N^T requires 2nd order quadrature, integrating B*B^T requires 1st order quadrature. And we only have elastic cases here, so 4-point quandrature is very sufficient.

I am not surprised by the changes. Even with the original implementation, mesh refinement would give an omega that is closer to what we have now, instead of the reference value.

For example, changing mesh size to 2 alone,

```py
rectangle_geometry.create_mesh(mesh_sizes=2)
```
leads to failure

```text
FAILURE: check 317541841.2903414 == pytest.approx(317723400.0, rel=1e-06, abs=None)
```

How the reference value was obtained? Is it numerial result or theoretical value?

robbievanleeuwen: Thanks for the refresher @TLCFEM, I am happy to keep the changes you have suggested. From memory the rectangular and angle results were obtained from Strand7, so they are numerical results. There was also no way to control the mesh size at the time so they are by no means refined. I no longer have access to Strand7 to confirm, but I am happy with the results as is :)

TLCFEM: I wanted to use 3-point quadrature but it failed so many tests. I do not have time to refactor those for the moment so I do not touch it. Those reference values may need to be updated by using a denser mesh, if they are numerical results.

For `fea.py`, there is nothing much can be done that may significantly optimise the performance for the moment. I'll check other files.

PR #300, opened 2023-10-03
Allow unlipped cee and zed steel_sections

smith120bh:
_This PR recreates #285 , but is based upon the `hypermodern` branch from the outset instead!_

--- Original description below ---
Unlipped Cee sections are very common. For example, all track ("T") sections in the Steel Framing Industry Association standard sections, [as defined here](https://sfia.memberclicks.net/assets/TechFiles/sfia%20tech%20spec%202015%20updated%207.12.17.pdf), are unlipped Cee sections. However, section-properties currently explicitly throws an error if lip length l is set to anything less than r_out (the outer radius).

This pull request adds the ability to set l to zero - or anything less than r_out.

Note that I did have a question as to what to do for 0 < l < r_out. One option would be to throw an error in this case, but I opted instead for decreasing the outer radius commensurately, on the philosophy that fewer error conditions is probably better, but am happy to change this as well.

While unlipped Zed sections are less common than unlipped Cees, they do exist - and a nearly-identical update could be applied to zed_section() to allow unlipped zeds as well here.

### DIFF
--- a/src/sectionproperties/pre/library/steel_sections.py
+++ b/src/sectionproperties/pre/library/steel_sections.py
@@ -1103,7 +1103,7 @@ def cee_section(
         points += sp_utils.draw_radius(
             pt=(b - t - r_in, t + r_in), r=r_in, theta=0, n=n_r, ccw=False
         )
-    
+
     # if the lip is shorter than the outer radius (curve only)
     elif l > t and l <= r_out:
         # construct a smaller corner for bottom right if t < l < r_out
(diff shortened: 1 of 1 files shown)


smith120bh:
> @robbievanleeuwen
[Blackify](https://github.com/robbievanleeuwen/section-properties/pull/300/commits/81d12769686308f88063c7600899130ac08ad02b)

Sorry about that! I missed that my black extension didn't format on save here.

robbievanleeuwen:
All good @smith120bh, contributor guide is pretty vague at the moment - once i get this big PR merged using poetry and pre-commit will make contributing much easier and standardised.

I've also just removed the `ValueError` thrown `if l < r_out` which I assume is no longer relevant?

I've had a play around with a number of edge cases and it seems to work well, thanks!

robbievanleeuwen (approved):

PR #324, opened 2023-10-05
Add numba, pardiso

TLCFEM:
1. Further jit shape function computation.
2. Provide alternative to use pardiso as sparse solver.

The default superlu solver is unbearably slow on large systems. With this PR, the warping computation is now ~10% of the time required by v2.1.5.

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -32,7 +32,7 @@ Installing ``PARDISO`` Solver
 
 The default sparse solver used in ``scipy`` is ``SuperLU``.
 It performs okay for small matrices but appears to be very slow for larger matrices.
-The ``PARDISO`` solver is a much faster alternative, but it requires the installation of the ``MKL`` library, which takes a lot of disk space.
+The ``PARDISO`` solver is a much faster alternative (see `pypadiso <https://github.com/haasad/PyPardisoProject>`_), but it requires the installation of the ``MKL`` library, which takes a lot of disk space.
 
 If you do not have a disk space constraint, you can use the ``PARDISO`` solver by:
 
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -26,6 +26,20 @@ package index:
 
     pip install sectionproperties
 
+
+Installing ``PARDISO`` Solver
+-----------------------------
+
+The default sparse solver used in ``scipy`` is ``SuperLU``.
+It performs okay for small matrices but appears to be very slow for larger matrices.
+The ``PARDISO`` solver is a much faster alternative, but it requires the installation of the ``MKL`` library, which takes a lot of disk space.
+
+If you do not have a disk space constraint, you can use the ``PARDISO`` solver by:
+
+.. code-block:: shell
+
+    pip install sectionproperties[pardiso]
+
 Installing CAD Modules
 ----------------------
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -32,7 +32,7 @@ classifiers = [
 ]
 packages = [
     { include = "sectionproperties", from = "src" },
-    { include = "src/sectionproperties/py.typed"},
+    { include = "src/sectionproperties/py.typed" },
 ]
 include = [
     "*.3dm",
@@ -48,7 +48,7 @@ Changelog = "https://github.com/robbievanleeuwen/section-properties/releases"
 
 [tool.poetry.dependencies]
 python = ">=3.9.0,<3.12"
-numpy = "^1.25.2"
+numpy = "^1.25.2" # numba requires numpy <1.26
 scipy = "^1.11.3"
 matplotlib = "^3.8.0"
 shapely = "^2.0.1"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -48,7 +48,7 @@ Changelog = "https://github.com/robbievanleeuwen/section-properties/releases"
 
 [tool.poetry.dependencies]
 python = ">=3.9.0,<3.12"
-numpy = "^1.26.0"
+numpy = "^1.25.2"
 scipy = "^1.11.3"
 matplotlib = "^3.8.0"
 shapely = "^2.0.1"
@@ -56,9 +56,11 @@ triangle = "^20230923"
 rich = "^13.6.0"
 click = "^8.1.7"
 more-itertools = "^10.1.0"
+numba = "^0.58.0"
 cad-to-shapely = { version = "^0.3.1", optional = true }
 rhino-shapley-interop = { version = "^0.0.4", python = ">=3.9.0,<3.10", optional = true }
 rhino3dm = { version = "==8.0.0b2", python = ">=3.9.0,<3.10", optional = true }
+pypardiso = { version = "^0.4.2", optional = true }
 
 [tool.poetry.dev-dependencies]
 black = "^23.9.1"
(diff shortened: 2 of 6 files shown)


src/sectionproperties/analysis/fea.py
```diff
@@ -12,12 +12,160 @@
...
 
 import numpy as np
+from numba import njit
 
 
 if TYPE_CHECKING:
     from sectionproperties.pre.pre import Material
 
 
+@njit(cache=True, nogil=True)
```
TLCFEM, line 22: Making those utilities pure numerical so that they can be jited.

src/sectionproperties/analysis/fea.py
```diff
@@ -902,6 +1047,61 @@ def shape_function_only(p: tuple[float, float, float]) -> np.ndarray:
     )
 
 
+h_inv = np.array(
```
TLCFEM, line 1050: Avoid repeatedly creating and destroying this constant matrix.

src/sectionproperties/analysis/section.py
```diff
@@ -454,9 +454,7 @@ def solve_warping() -> np.ndarray:
             else:
                 ixx_c, iyy_c, ixy_c = self.get_ic()
 
-            self.section_props.j = (
-                ixx_c + iyy_c - omega.dot(k_lg[:-1, :-1].dot(np.transpose(omega)))
```
TLCFEM, line 458: Overcomplex computation as U is the solution to KU=F

TLCFEM:
@robbievanleeuwen Should be ready, please check it at your convenience.

robbievanleeuwen:
Thanks again for your hard work @TLCFEM - I will try to have a look at this in the next few days :)

pyproject.toml
```diff
@@ -48,17 +48,19 @@ Changelog = "https://github.com/robbievanleeuwen/section-properties/releases"
 
 [tool.poetry.dependencies]
 python = ">=3.9.0,<3.12"
-numpy = "^1.26.0"
+numpy = "^1.25.2"
```
robbievanleeuwen, line 51: Could you put a comment here stating the reason for the downgrade so I can remember to check this if `numba` allows further `numpy` updates.

```
e.g. numpy = "^1.25.2"  # numba requires numpy <1.26
```

pyproject.toml
```diff
@@ -48,17 +48,19 @@ Changelog = "https://github.com/robbievanleeuwen/section-properties/releases"
...
 shapely = "^2.0.1"
 triangle = "^20230923"
 rich = "^13.6.0"
 click = "^8.1.7"
 more-itertools = "^10.1.0"
+numba = "^0.58.0"
 cad-to-shapely = { version = "^0.3.1", optional = true }
 rhino-shapley-interop = { version = "^0.0.4", python = ">=3.9.0,<3.10", optional = true }
 rhino3dm = { version = "==8.0.0b2", python = ">=3.9.0,<3.10", optional = true }
+pypardiso = { version = "^0.4.2", optional = true }
```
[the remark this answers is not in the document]
robbievanleeuwen: I agree that it is a good idea to make it optional, making it a core dependency will also mean that it needs to be downloaded and installed for CI (e.g. tests, docs build).

docs/installation.rst
```diff
@@ -26,6 +26,20 @@ package index:
 
     pip install sectionproperties
 
+
+Installing ``PARDISO`` Solver
+-----------------------------
+
+The default sparse solver used in ``scipy`` is ``SuperLU``.
+It performs okay for small matrices but appears to be very slow for larger matrices.
+The ``PARDISO`` solver is a much faster alternative, but it requires the installation of the ``MKL`` library, which takes a lot of disk space.
```
robbievanleeuwen, line 35: This section reads really well, thanks! I would just suggest providing a link the the pypardiso GitHub page so users can easily decide whether or not they would like to install this. For example:

> The ``PARDISO`` solver is a much faster alternative (see [PyPardiso](https://github.com/haasad/PyPardisoProject)), but it requires...

src/sectionproperties/analysis/fea.py
```diff
@@ -420,17 +566,19 @@ def element_stress(
         # calculate axial stress
         sig_zz_n = n * np.ones(6) * self.material.elastic_modulus / ea
```
robbievanleeuwen, line 567: Probably make this ``np.ones(n_points)`` for consistency.

src/sectionproperties/analysis/section.py
```diff
@@ -131,21 +131,8 @@ def __init__(
...
-            x4 = self._mesh_nodes[node_ids[3]][0]
-            y4 = self._mesh_nodes[node_ids[3]][1]
-            x5 = self._mesh_nodes[node_ids[4]][0]
-            y5 = self._mesh_nodes[node_ids[4]][1]
-            x6 = self._mesh_nodes[node_ids[5]][0]
-            y6 = self._mesh_nodes[node_ids[5]][1]
-
             # create a list containing the vertex and mid-node coordinates
-            coords = np.array([[x1, x2, x3, x4, x5, x6], [y1, y2, y3, y4, y5, y6]])
+            coords = self._mesh_nodes[node_ids, :].transpose()
```
robbievanleeuwen, line 135: Nice!

robbievanleeuwen (changes requested):
This looks really good @TLCFEM, thanks again!

Running the newly added benchmarks, I got a significantly faster runtime for the heaviest warping analysis on my machine with the `paradiso` solver (4.6 secs vs. 1.6secs), well done!

The only thing apart from the minor comments is a warning that is printed when running the tests:

```
tests/analysis/test_stress.py::test_stress_runtime_errors
  /home/runner/work/section-properties/section-properties/src/sectionproperties/analysis/fea.py:1114: NumbaPerformanceWarning: np.dot() is faster on contiguous arrays, called on (Array(float64, 2, 'C', True, aligned=True), Array(float64, 1, 'A', False, aligned=True))
    return h_inv.dot(w)
```

I'm never used `numba` before so not sure if this is worth addressing?

TLCFEM:
> This looks really good @TLCFEM, thanks again!
> 
> Running the newly added benchmarks, I got a significantly faster runtime for the heaviest warping analysis on my machine with the `paradiso` solver (4.6 secs vs. 1.6secs), well done!
> 
> The only thing apart from the minor comments is a warning that is printed when running the tests:
> 
> ```
> tests/analysis/test_stress.py::test_stress_runtime_errors
>   /home/runner/work/section-properties/section-properties/src/sectionproperties/analysis/fea.py:1114: NumbaPerformanceWarning: np.dot() is faster on contiguous arrays, called on (Array(float64, 2, 'C', True, aligned=True), Array(float64, 1, 'A', False, aligned=True))
>     return h_inv.dot(w)
> ```
> 
> I'm never used `numba` before so not sure if this is worth addressing?

Not a great concern as this omega is passed in as a view of some other matrix, which happen to be row major that causes non-contiguous arrays. To suppress this warning, either transpose the following:

```py
        sig_zxy_mzz_gp = np.zeros((n_points, 2))
        sig_zxy_vx_gp = np.zeros((n_points, 2))
        sig_zxy_vy_gp = np.zeros((n_points, 2))
```

or replace dot with matmul. I chose the latter.

robbievanleeuwen (approved):

Issue #331, opened 2023-10-09
`CompoundGeometry.offset_perimeter()` with positive `amount` creates overlapping geometry


robbievanleeuwen:
See below example:

```python
from sectionproperties.pre.geometry import check_geometry_overlaps
from sectionproperties.pre.library import rectangular_section
from sectionproperties.analysis import Section


# create two rectangles side-by-side
rect1 = rectangular_section(d=50, b=50)
rect2 = rectangular_section(d=50, b=50).align_to(rect1, "right")
geom = rect1 + rect2

# dilate the exterior by 5 units
geom = geom.offset_perimeter(amount=5, where="exterior")

# ensure there are no overlaps
print(check_geometry_overlaps([g.geom for g in geom.geoms]))

# plot mesh
geom.create_mesh(mesh_sizes=[0])
Section(geometry=geom).plot_mesh(materials=False)
```

`check_geometry_overlaps()` returns `True` - this should be `False`.

Mesh plot shows two semi-circles overlapping the existing rectangles in the regions where the two rectangles meet:
![Screenshot 2023-10-09 at 10 08 26 pm](https://github.com/robbievanleeuwen/section-properties/assets/18841910/c04850b0-5994-4a1a-be9c-2f75fb220a02)

connorferster:
Not quite sure how to fix this yet. It's a tricky one.

PR #332, opened 2023-10-09
Fix `CompoundGeometry` offset perimeter (dilation creates overlapping geometry)

robbievanleeuwen:
This PR addresses #331, in which dilated offsets create overlapping geometry.

Tasks:
- [x] Successfully run `test_compound_rectangular_offset()`


robbievanleeuwen:
@connorferster - seeing as you wrote the original code for this, wondering if you have time to have a quick look?

connorferster:
@robbievanleeuwen For your review. The tests pass but mypy does not like my type annotations on one helper function. I am unsure of how to interpret the error messages in order to fix. 

The function takes input the looks like this:

```python
array([
    <GEOMETRYCOLLECTION (MULTILINESTRING EMPTY, MULTILINESTRING ((500 500, 500 0)))>,
    <GEOMETRYCOLLECTION (MULTILINESTRING EMPTY, MULTILINESTRING ((1000 0, 1000 5...>],
     dtype=object
)
```

I originally tried `npt.ArrayLike[GeometryCollection]` but then switched it to `npt.ArrayLike` because, from the error messages, it seemed to be the thing to do.

Any insight on what to be done to get this merged?

connorferster:
Also, for fun: 

The live-stream for doing this PR: 
https://www.youtube.com/live/hSfsojAAJjc?feature=shared

robbievanleeuwen:
Thoroughly enjoyed the stream - helps with the code review as well 😆 Shapely is such a fun and feature rich library!

`mypy` - happy to ignore these two lines, in this case `shapely` isn't being particularly helpful with what it's returning and I'm happy with what you're doing here.

I've also added that test you were working on in the live stream with 3 rectangles!

Thanks again for working on this one, bit of a curly one in the end! I also wouldn't lose too much sleep on all the edge cases as I'm sure there are many, and I've got a little [note](https://sectionproperties.readthedocs.io/en/stable/user_guide/geometry.html#id36) in the docs mentioning that the user should double check the output geometry in these cases!

PR #345, opened 2023-10-13
Plot warping function only over section extents

robbievanleeuwen:
@TLCFEM - I made a few change to the warping function plot so that it only displays over the extent of the section. Do you think this is a better implementation? Consider this example:

```python
from sectionproperties.pre.library import channel_section
from sectionproperties.analysis import Section
geom = channel_section(200, 100, 16, 8, 15, 8)
geom.create_mesh(3)
sec = Section(geom)
sec.calculate_frame_properties()
sec.plot_warping_function()
```

With the current implementation you get this:

This PR gives:

I feel like this is a better visualisation, however would you ever want to visualise the entire field, even outside the section extents?

### DIFF
--- a/src/sectionproperties/analysis/section.py
+++ b/src/sectionproperties/analysis/section.py
@@ -12,6 +12,7 @@ from typing import Callable
 
 import matplotlib.axes
 import matplotlib.patches as mpatches
+import matplotlib.tri as tri
 import numpy as np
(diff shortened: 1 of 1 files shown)


TLCFEM:
Yes, this is for sure the more correct solution. I was not thoughtful and just wanted to plot something to show the difference.

PR #349, opened 2023-10-20
Make `numba` an optional dependency

### LINKED ISSUE #348: Section-properties is not running for me using Python 3.11.6
### Discussed in https://github.com/robbievanleeuwen/section-properties/discussions/347

<div type='discussions-op-text'>

<sup>Originally posted by **aladdin100001** October 19, 2023</sup>
I am not sure what is the problem and I hope I can get some guidance. I downloaded section-properties four months ago and it was working for me using python 3.9 (I am using visual studio IDE). I downloaded the latest section-properties yesterday and it is not working for me using python 3.9 or 3.11. Does anyone experience the same problem? </div>

Make ``numba`` an optional dependency for `sectionproperties` as some users have difficulties with the `numba` package.

robbievanleeuwen:
Fixes #348.

### DIFF
--- a/src/sectionproperties/analysis/fea.py
+++ b/src/sectionproperties/analysis/fea.py
@@ -10,19 +10,63 @@ from __future__ import annotations
 import warnings
 from dataclasses import dataclass, field
 from functools import lru_cache
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Any, Callable
 
 import numpy as np
 import numpy.typing as npt
-from numba import njit
-from numba.core.errors import NumbaPerformanceWarning
 
 
 if TYPE_CHECKING:
     from sectionproperties.pre.pre import Material
 
 
-@njit(cache=True, nogil=True)  # type: ignore
+# numba is an optional dependency
+try:
+    from numba import njit
+    from numba.core.errors import NumbaPerformanceWarning
+
+    USE_NUMBA = True
+except ImportError:
+
+    def njit() -> None:
+        """Assigns empty function to njit if numba isn't installed."""
+        return None
+
+    USE_NUMBA = False
+
+
+def conditional_decorator(
+    dec: Callable[[Any], Any],
+    condition: bool,
+) -> Callable[[Any], Any]:
+    """A decorator that applies a decorator only if a condition is True.
+
+    Args:
+        dec: Decorator to apply
+        condition: Apply decorator if this is true
+
+    Returns:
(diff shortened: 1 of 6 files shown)


src/sectionproperties/analysis/fea.py
```diff
@@ -55,7 +103,7 @@ def _assemble_torsion(
     return k_el, f_el, c_el
 
 
-@njit(cache=True, nogil=True)  # type: ignore
```
TLCFEM, line 58: This impairs performance as compiled code is not cached. Why not directly define a njit decorator when njit cannot be imported. This solution appears to be overcomplex to me.

src/sectionproperties/analysis/fea.py
```diff
@@ -648,7 +696,9 @@ def element_stress(
 
         # extrapolate results to nodes, ignore numba warnings about performance
         with warnings.catch_warnings():
-            warnings.simplefilter("ignore", category=NumbaPerformanceWarning)
+            if USE_NUMBA:
+                warnings.simplefilter("ignore", category=NumbaPerformanceWarning)
```
TLCFEM, line 700: Change line 638 to:

```py
sig_zxy_mzz_gp = np.zeros((n_points, 2), order='F')        sig_zxy_vx_gp = np.zeros((n_points, 2), order='F')
sig_zxy_vy_gp = np.zeros((n_points, 2), order='F')
```

can suppress the warning so no need to import those extra utilities.

src/sectionproperties/analysis/fea.py
robbievanleeuwen: Now that you mention this, that's an obvious solution thanks! I didn't realise this would impair the caching. If you have time to implement a fix I'd be happy to merge, otherwise I will get to this next week. Cheers!

TLCFEM: I mean, directly define the decorator on line 31:

```py
def njit(**options):
    def decorator(func):
        def wrapper(*args, **kwargs):
            return func(*args, **kwargs)

        return wrapper

    return decorator
```

TLCFEM: The rest does not need to be changed, so cache will still work if numba is installed. Ppl download and run once, the functions will be compiled, cached, and reused. Without cache, those functions will be compiled every time when the analysis is performed. Since most ppl do not change the source code, caching could benefit most users I presume.

PR #439, opened 2024-06-11
Update dependencies to swap in cytriangle

connorferster:
# Pull Request Check List

- [x] Added **tests** for changed code.
- [ ] Updated **documentation** for changed code.
- [x] Run the **Nox** test suite to check for errors and warnings.
- [ ] Check benchmarks `triangle` vs `cytriangle`


m-clare:
Patch release [v1.0.1](https://github.com/m-clare/cytriangle/releases/tag/v1.0.1) adds `macos-13` builds to PyPi. 

I can add typing this weekend to appease mypy if necessary.

robbievanleeuwen:
Thanks @m-clare, no need to add typing, I excluded `triangle` from mypy previously and am happy to exlude `cytriangle` which now fixes those workflows.

@connorferster I have a few weeks off work starting next week so am happy to update to update the docs and merge.

Also, we are now testing against `macos-14` (i.e. arm64), yay 🚀 Note that `rhino3dm` has been updated to suit and the tests pass, however I have issues with installing recent versions of `rhino3dm` on my mac at home, see mcneel/rhino3dm#622. I think there is an issue with the universal wheels but not 100% sure? Anyway I am happy to merge this branch for now, if a user experiences this it's a pretty easy fix by manually downloading the appropriate `rhino3dm` wheel from pypi and installing into their virtual environment.

connorferster:
Hi @robbievanleeuwen,

I will merge this but then also ask: how do I check the benchmarks?

robbievanleeuwen:
Thanks @connorferster, there's a bit of info [here](https://sectionproperties.readthedocs.io/en/stable/contributing.html#benchmarks) about the benchmarks. Basically just runs a set of tests and gives some time info as output. I can have a look at these tomorrow, will just checkout the code before and after the PR merge to see if there's any significant changes. I'll also quickly update the docs where it references the old triangle library - should be pretty quick with a little `ctrl-f` action!

robbievanleeuwen:
# Benchmarking

These results are in no way scientific (run on a linux virtual machine with background tasks going running on my PC):

## Meshing

`triangle` appears to be a fair bit faster than `cytriangle`, but note the results are in microseconds so don't make a huge difference to the runtime of `sectionproperties. The difference between the two libraries appears to grow as mesh size grows so maybe something to investigate if you're interested in optimisation? Again, take these numbers with a grain of salt as I'm no expert at benchmarking...

### Cytriangle

mesh_cytriangle

### Triangle

mesh_triangle

## Analysis

This is a benchmark of the whole `sectionproperties` workflow. The `cytriangle` runs are faster than the `triangle` runs, I can't explain this. So maybe we throw everything out the window 😆 

### Cytriangle

analysis_cytriangle

### Triangle

analysis_triangle

PR #466, opened 2024-10-12
Add rectangular CLT section to (new) timber sections library

jchkoch:
# Feature Timber Sections

Discussion: #437 

This is a start at adding timber cross-section analysis. Please let me know what you think of the approach to add timber sections. So far I have added geometry creation of layered timber sections. I have added tests for the code that I have written so far. 

Moving forward I will be checking and validating the results from the cross-sectional properties calculated for the sections as well as the stress analysis modules.

- [X] Added **tests** for changed code.
- [x] Updated **documentation** for changed code.
- [x] Run the **Nox** test suite to check for errors and warnings.


robbievanleeuwen:
Looks promising! I think this would be a worthy addition to `sectionproperties`.

I've also included @connorferster as a reviewer as I believe he has more timber expertise than myself!

Looking forward to reviewing the complete PR!

FYI - the docs for this PR can be found [here](https://sectionproperties--466.org.readthedocs.build/en/466/).

connorferster:
I think it looks good! @jchkoch Glad you dropped the timber rectangular section in favour of the CLT section. Setting the shear modulus of the cross-lams via the `elastic_modulus` and `poissons_ratio` is the way to do it but it sure feels weird to see it (just a general comment, not a reason to change anything).

If we can get the tests passing, then I am good to go on this one.

src/sectionproperties/pre/library/timber_sections.py
```diff
@@ -0,0 +1,66 @@
...
+    Example:
+        The following example creates a 120mm CLT cross-section.
+    """
+
+    layer_geom = list()
+    for idx in range(len(d)):
+        di = float(d[idx])
+        layer = lay_orient[idx]
+
+        if layer is int(0):
```
connorferster, line 47: This seems redundant and misses the mark slightly. Could it not just be `if layer == 0`?

src/sectionproperties/pre/library/timber_sections.py
```diff
@@ -0,0 +1,66 @@
...
+) -> geometry.CompoundGeometry:
+    """Constructs a timber rectangular section.
+
+    Constructs a timber rectangular section of depth ``d`` and width ``b``.
+
+    .. note::
+
+    Args:
+        d: Timber layer section thickness
+        lay_orient: List of layer orientation
```
connorferster, line 25: I think `lay_orient` needs a little more explanation. The implementation of this function assumes certain values to be in there, namely `0` and any other value. The lay orientation could just as well be `[0, 1, 0, 1, 0]` instead of `[0, 90, 0, 90, 0]` and the function would work the same (it could also use `[0, 34, 0, 
-15, 0,]` and you would get the same result).

I think a list of values to describe the layup makes sense but it seems to rely, somewhat, on "magic values" that are not explained. I think it would be good to say something like:

`lay_orient: A list of layer orientations, e.g. [0, 90, 0, 90, 0], where 0 indicates where `timb_mat0` is to be used and where 90 indicates where `timb_mat1` is to be used.`

jchkoch:
@connorferster Thanks for the comments. Unfortunately, I became somewhat busy and haven't managed to follow up.  However, I should have a bit more time going forward.

robbievanleeuwen:
Thanks for checking in @jchkoch. If you don't mind I can update your feature branch to the current state of `sectionproperties`, in which the CI/CD has changed a little bit since this PR was opened (not sure if you can see the "Update branch" button as well - happy for you to click this if you can see it!)

src/sectionproperties/pre/library/timber_sections.py
jchkoch: Thanks, for the comment, I have adjusted this as suggested.

src/sectionproperties/pre/library/timber_sections.py
jchkoch: That makes sense. I have updated the code as discussed.

jchkoch:
Hi @robbievanleeuwen, I have updated the pull request as discussed above to address @connorferster comments as well as merging the commits from the main branch into my feature branch. I have also updated the docs with the new function `clt_rectangular_section` under section.rst. The docs seem to be able to build properly.

However, when validating the results of a geometric analysis (in the example jupyter notebook `section_library.ipynb`) of a CLT section the moment of inertia of the minor axis (Iyy) seems to not be correct. The major axis is correct. I have not been able to find the error so far. Will try again in the next few days.

robbievanleeuwen:
Hi @jchkoch, thanks again for your hard work on this. I should have some time at the end of this week to review. Did you manage to get to the bottom of the validation issues? If not I can try figure it out as well :)

jchkoch:
I have not been able to figure out the validation issue up to now. I thought I would leave it and come back to it. I may also have a bit more time in the next few days to look at it again.

.python-version
robbievanleeuwen, line 1: Could you please add this file back to the branch?

docs/examples/geometry/section_library.ipynb
robbievanleeuwen, line 1: This looks good! I think you have accidentally copied a couple of headings here by mistake - the cells with "## Rectangular Concrete Section", "### Import Modules" and "from sectionproperties.pre import Material" (cells 37 to 39) should be deleted.

Also could you place the "Rectangular Timber CLT Section" before the "Rectangular Concrete Section" example? By default the examples index [page](https://sectionproperties.readthedocs.io/en/stable/examples.html) uses the last image as a thumbnail and the CLT image is a funny aspect ratio to be the thumbnail. 

Also in the last few cells there are references to "concrete", which should be timber or CLT. And the text mentions comparing the moment of inertia to a rectangle, however in the example you are calculating x and y moments of inertia.

Also I've realised in the reinforced concrete example I use the units "mm6", but these should be "mm4" - would you be able to fix up this reference for me and also the units in your example? Thanks :)

docs/user_guide/geometry.rst
robbievanleeuwen, line 1: Would you be ble to move this above the label the bridge library (.. _label-bridge-library:) - this is a cross reference from another page which needs to reference the Bridge Sections heading.

src/sectionproperties/pre/library/timber_sections.py
```diff
@@ -0,0 +1,54 @@
...
+    Raises:
+        ValueError: Geometry generation failed
+
+    Returns:
+        Timber rectangular section geometry
+
+    Example:
+        The following example creates a 120mm CLT cross-section.
+    """
+    layer_geom = []
```
robbievanleeuwen, line 33: To fix type checking this can be `layer_geom: list[geometry.Geometry] = []`

src/sectionproperties/pre/library/timber_sections.py
```diff
@@ -0,0 +1,54 @@
...
+
+        # create rectangular timber geometry
+        layer = primitive_sections.rectangular_section(d=di, b=b, material=timb_mat)
+        offset = -d[idx] * (idx + 1)
+        layer = layer.shift_section(y_offset=offset)
+
+        layer_geom.append(layer)
+
+    # create compound geometry
+    geom = geometry.CompoundGeometry(geoms=layer_geom)
```
robbievanleeuwen, line 48: This can be `return geometry.CompoundGeometry(geoms=layer_geom)`. Everything below this can be deleted (fixes type checking, also not required in this case - other functions need it as we are building compound geometry through boolean operations however this function just creates a `CompoundGeometry` object so we'll always have a `CompoundGeometry` object!)

robbievanleeuwen (commented):
I've added a few small code/docs comments. Otherwise looking great! Thanks for adhering to the style and adding tests/docs!

Tomorrow I'll have a look at why the validation example isn't working as expected.

src/sectionproperties/pre/library/timber_sections.py
```diff
@@ -0,0 +1,54 @@
...
+) -> geometry.CompoundGeometry:
+    """Constructs a timber rectangular section.
+
+    Constructs a timber rectangular section of depth ``d`` and width ``b``.
+
+    .. note::
+
+    Args:
+        d: Timber layer section thickness
+        lay_orient: A list of materials for each layer defined by the user.
```
robbievanleeuwen, line 21: It may be worth adding that layers are added from top to bottom in this implementation (users may expect a bottom up by default).

jchkoch:
Hi @robbievanleeuwen, validation seems to work. I think I managed to confuse myself with regard to the coordinate system used by section-properties. I was indeed trying to validate the second case for bending where you have the timber90 at the top and bottom and timber0 in the center. From a practical point of view for 2-way CLT design, the so-called major and minor plate bending directions are particularly interesting. Making a new geometry object works!

Major (x-) axis bending (Timber0 at the top and bottom)
I_eff,x = 138.8e6 mm^4
Matches output from section-properties

Minor (z-) axis bending (Timber90 at the top and bottom)
I_eff,y = 9.960e6 mm^4
Matches output from section-properties

.python-version
jchkoch: Added file back. Not sure what happened here.

docs/user_guide/geometry.rst
jchkoch: Fixed.

src/sectionproperties/pre/library/timber_sections.py
jchkoch: Fixed.

src/sectionproperties/pre/library/timber_sections.py
jchkoch: Fixed.

robbievanleeuwen (approved):

robbievanleeuwen:
Thanks for your hard work on this @jchkoch, looks great!

Issue #572, opened 2025-11-12
Support python 3.14


SZeltaat:
- [x] I have searched the [issues](https://github.com/robbievanleeuwen/section-properties/issues) of this repo and believe that this is not a duplicate.
- [x] I have searched the [documentation](https://sectionproperties.readthedocs.io/) and believe that my question is not already covered.

## Feature Request
Hi 👋,

First, thank you for maintaining section-properties. It’s a great tool and an important dependency for our project.

We recently updated our own package to support Python 3.14, but our workflows are now failing because section-properties specifies an upper Python version bound (<3.14). This introduces problems in solving the dependency graph.

Would it be possible to update the supported Python versions to include 3.14?

If time is a constraint, we’d be happy to help by testing the package under Python 3.14 and contributing a pull request to update compatibility.

Thanks again for your work on this project!

PR #573, opened 2025-12-01
Support python 3.14

connorferster:
# Pull Request Check List

Resolves: #572 

- [x] Added a description of your pull request below.
- [x] Added **tests** for changed code.
- [x] Updated **documentation** for changed code.
- [x] Run `uv run pre-commit run --all-files` and `uv run pyright` to ensure code quality.

Update project to support Python 3.14

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,7 +9,7 @@ on:
 
 env:
   UV_VERSION: 0.9.14
-  DEFAULT_PYTHON_VERSION: '3.14'
+  DEFAULT_PYTHON_VERSION: '3.13'
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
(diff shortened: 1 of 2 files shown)


robbievanleeuwen:
Hi @connorferster thanks for picking this one up, unfortunately I've been too busy to spend any time on `sectionproperties` lately, hoping this will change soon!

I am happy for the upgrade to 3.14 to go ahead while leaving `rhino3dm` behind, I have done this before. If we get to the release point and this is still the case, we can add a little note on the [installation page](https://sectionproperties.readthedocs.io/en/stable/installation.html#installing-cad-modules). I agree that it would be good to wait until the imminent release of `numba` as I believe that this is used by a number of users.

I have rolled back the read the docs build to use python 3.13 as this uses rhino for an example and don't believe we need the cutting edge to build the docs (or hold back the upgrade).

